PR #3, opened 2020-01-11
Merge ShooterSubsystem into Master

FireShredder24:
The ShooterSubsystem provides the methods necessary to control the NEO Brushless Motor that will be mounted on the Power Cell launcher.


src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -7,46 +7,93 @@
...
+    // from 0 speed.
+    ShooterMAX.setIdleMode(IdleMode.kCoast);
   }
 
+  /**
+   * Passes a preset velocity to the SparkMAX PIDF controller and lets it manage
+   * the NEO's velocity. Intended to be called when a button is pressed.
+   */
   public void shoot() {
     ShooterPID.setReference(ShooterSetpoint, ControlType.kVelocity);
```
jordan-powers: Would it be more advantageous to use a parameter on the shoot() method for the shoot speed? That way we might be able to shoot the ball different distances

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -7,46 +7,93 @@
...
+import com.revrobotics.CANSparkMax.IdleMode;
 
 import frc.robot.Constants;
 
 public class ShooterSubsystem extends SubsystemBase {
+  /**
+   * Initializes the SparkMAX motor controller, assigns it to the CAN address
+   * specified, and sets it to the NEO Brushless Motor.
+   */
   private final CANSparkMax ShooterMAX = new CANSparkMax(Constants.SPARKMAX_SHOOTER_CAN_ADDR,
```
jordan-powers: Java language conventions dictate that class names start with a capital letter and member variables and methods start with a lower case letter
https://docs.oracle.com/javase/tutorial/java/nutsandbolts/variables.html

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -7,46 +7,93 @@
...
       CANSparkMaxLowLevel.MotorType.kBrushless);
-
+  /**
+   * The built-in PID controller provided by the Spark MAX motor controller.
+   */
   private final CANPIDController ShooterPID = new CANPIDController(ShooterMAX);
```
jordan-powers: Lowercase s

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -7,46 +7,93 @@
...
-
-  private final Encoder shootEncoder = new Encoder(Constants.SHOOTER_ENCODER_CHAN_A, Constants.SHOOTER_ENCODER_CHAN_B);
-
+  /**
+   * The target velocity of the NEO Brushless Motor.
+   */
   private final double ShooterSetpoint = 1.0;
```
jordan-powers: Lowercase s

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -7,46 +7,93 @@
...
   private final double kIz = 1;
+  /**
+   * The Feed-Forward Gain of the SparkMAX PIDF controller. The weight of the
+   * feed-forward loop as compared to the PID loop is controlled by this value.
+   */
   private final double kFF = 1;
+  /**
+   * Scales the output of the SparkMAX PIDF controller.
+   */
   private final double OutputRange = 1;
```
jordan-powers: Lowercase o

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
FireShredder24: That's what the "hood" is for.  Check @design_team and look at Jake's CAD for what I mean.  There will be a separate subsystem for control of the hood.

jordan-powers:
It all looks fine to me. I don't know if @terbos90803 wants to review it, but I don't have any problems with merging this to master.

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -7,44 +7,91 @@
...
+  /**
+   * The target velocity of the NEO Brushless Motor.
+   */
+  private final double shooterSetpoint = 1.0;
+  /**
+   * The Proportial Gain of the SparkMAX PIDF controller The weight of the
+   * proportional path against the differential and integral paths is controlled
+   * by this value.
+   */
   private final double kP = 1;
```
terbos90803: Research default PID constants for the NEO.  You'll find that they're much smaller than 1.  Setting them to 1 can be surprisingly violent.  (Ask Jake about his laptop)

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -7,44 +7,91 @@
...
+  /**
+   * The Integral Gain of the SparkMAX PIDF controller The weight of the integral
+   * path against the proportional and differential paths is controlled by this
+   * value.
+   */
   private final double kI = 1;
```
terbos90803: Default constants for I, D, FF, etc. should be 0 or set from known defaults.  Never default these to 1 as placeholders.

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -7,44 +7,91 @@
...
-    ShooterPID.setIZone(kIz, 0);
-    ShooterPID.setFF(kFF, 0);
-    ShooterPID.setOutputRange(-OutputRange, OutputRange, 0);
+    // Applies the previously-declared values to the PIDF controller.
+    shooterPID.setP(kP, 0);
+    shooterPID.setI(kI, 0);
+    shooterPID.setD(kD, 0);
+    shooterPID.setIZone(kIz, 0);
+    shooterPID.setFF(kFF, 0);
+    shooterPID.setOutputRange(-outputRange, outputRange, 0);
```
terbos90803: Consider also setting the current limiting feature of the SparkMAX.  NEO should be 40A.

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -7,44 +7,91 @@
...
-  private final double ShooterSetpoint = 1.0;
-
+  /**
+   * The built-in PID controller provided by the Spark MAX motor controller.
+   */
+  private final CANPIDController shooterPID = new CANPIDController(shooterMAX);
+  /**
+   * The target velocity of the NEO Brushless Motor.
+   */
+  private final double shooterSetpoint = 1.0;
```
terbos90803: Expose this setpoint so it can be controlled.  A good first step is using MoPrefs.  If it needs to change too much, then break out as a Shuffleboard control.
It might also depend on the range calculation from the LimeLight.

terbos90803 (changes requested):

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -31,39 +31,44 @@
   /**
    * The target velocity of the NEO Brushless Motor.
    */
-  private final double shooterSetpoint = 1.0;
+  public final double shooterSetpoint = 1.0;
```
terbos90803: Why did you make this public?  Member variables should *never* be public and should *always* be modified via an accessor (set) method.

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -31,39 +31,44 @@
...
   /**
    * The Proportial Gain of the SparkMAX PIDF controller The weight of the
    * proportional path against the differential and integral paths is controlled
    * by this value.
    */
-  private final double kP = 1;
+  private final double kP = 0.1;
```
terbos90803: Still too big.  Review the Spark/NEO PID constants from last year or from the Rev sample code.  You'll find that P should be much smaller.

terbos90803 (changes requested):

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
FireShredder24: Done.

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
FireShredder24: Done

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -72,32 +75,46 @@
...
+    shooterPIDRight.setFF(kFF, 0);
+    shooterPIDLeft.setOutputRange(-outputRange, outputRange, 0);
+    shooterPIDRight.setOutputRange(-outputRange, outputRange, 0);
     // Sets the shooter motor to coast so that subsequent shots don't have to rev up
     // from 0 speed.
-    shooterMAX.setIdleMode(IdleMode.kCoast);
-    shooterMAX.setSmartCurrentLimit(currentLimit);
+    shooterMAXLeft.setIdleMode(IdleMode.kCoast);
+    shooterMAXRight.setIdleMode(IdleMode.kCoast);
+    shooterMAXLeft.setSmartCurrentLimit(currentLimit);
```
terbos90803, line 94: Because these two motors are directly connected together, they cannot be controlled independently or their PIDs will fight.  One controller must be selected as the primary and the other configured as a follower.

This is the same situation as with the Falcons on the drive base.

terbos90803 (changes requested):

terbos90803 (approved):

PR #4, opened 2020-01-14
Add GameData functionality to TeleopPeriodic

FireShredder24:
Adds functionality to get the Position Control color from the Driver Station to frc.robot.Robot.teleopPeriodic();

### DIFF
--- a/src/main/java/frc/robot/Robot.java
+++ b/src/main/java/frc/robot/Robot.java
@@ -10,6 +10,7 @@ package frc.robot;
 import edu.wpi.first.wpilibj.TimedRobot;
 import edu.wpi.first.wpilibj2.command.Command;
 import edu.wpi.first.wpilibj2.command.CommandScheduler;
+import edu.wpi.first.wpilibj.DriverStation;
 
 /**
  * The VM is configured to automatically run this class, and to call the
@@ -23,6 +24,8 @@ public class Robot extends TimedRobot {
   private Command m_autonomousCommand;
 
   private RobotContainer m_robotContainer;
+
+  private char positionControlColor;
 
   /**
    * This function is run when the robot is first started up and should be used
@@ -105,6 +108,28 @@ public class Robot extends TimedRobot {
    */
   @Override
   public void teleopPeriodic() {
+    String gameData;
+    gameData = DriverStation.getInstance().getGameSpecificMessage();
+    if (gameData.length() > 0) {
+      switch (gameData.charAt(0)) {
+      case 'B':
+        positionControlColor = 'B';
+        break;
+      case 'R':
+        positionControlColor = 'R';
+        break;
+      case 'Y':
+        positionControlColor = 'Y';
+        break;
+      case 'G':
+        positionControlColor = 'G';
+        break;
+      default:
+        break;
+      }
+    } else {
+
+    }
   }
 
   @Override


src/main/java/frc/robot/Robot.java
```diff
@@ -105,6 +108,28 @@ public void teleopInit() {
...
+        positionControlColor = 'Y';
+        break;
+      case 'G':
+        positionControlColor = 'G';
+        break;
+      default:
+        break;
+      }
+    } else {
+
```
jordan-powers, line 131: If you're not going to do anything in the `else` statement, then why include it?
This will only ever run if we receive an empty string from the FMS (aka when there's no data available).

src/main/java/frc/robot/Robot.java
```diff
@@ -105,6 +108,28 @@ public void teleopInit() {
...
+      case 'R':
+        positionControlColor = 'R';
+        break;
+      case 'Y':
+        positionControlColor = 'Y';
+        break;
+      case 'G':
+        positionControlColor = 'G';
+        break;
+      default:
```
jordan-powers, line 127: If we ever reach this default block, there's something wrong with data sent from the FMS so maybe print an error to the console.

src/main/java/frc/robot/Robot.java
```diff
@@ -24,6 +25,8 @@
 
   private RobotContainer m_robotContainer;
 
+  private char positionControlColor;
```
jordan-powers, line 28: This is probably not the right place for this.

A good principle of object-oriented programming is to try to reduce [coupling](https://en.wikipedia.org/wiki/Coupling_(computer_programming)) between the different parts of the code. This means that every class should know as little as possible about the other classes.

In regards to the requested color wheel position, that information is not needed by every part of the robot. For example, neither the drive nor the shooter care what the position of the color wheel should be. The only time we need to know the desired position of the color wheel is when we're actually setting the color wheel. So this logic should be part of the (as of yet unimplemented) `SetColorWheel` command, not part of the entire robot.

jordan-powers (changes requested):
My largest concern is that `Robot.java` is not the correct place to be interpreting the color wheel data.

src/main/java/frc/robot/Robot.java
terbos90803: It's going to be common to receive an empty or null string from FMS, so that does need to be handled gracefully.  But I agree that an empty else clause should be removed.

src/main/java/frc/robot/Robot.java
terbos90803: Based on the experience from 2018, it was common to get a null String.  That should be handled first.  The default clause would be a good place to log any unusual string contents.

src/main/java/frc/robot/Robot.java
```diff
@@ -105,6 +108,28 @@ public void teleopInit() {
    */
   @Override
   public void teleopPeriodic() {
+    String gameData;
+    gameData = DriverStation.getInstance().getGameSpecificMessage();
+    if (gameData.length() > 0) {
```
terbos90803, line 113: Needs a null check and can be written more clearly as:
`if (!(gameData == null || gameData.isEmpty()) {`

terbos90803 (changes requested):

FireShredder24:
Since mechanical decisions have now been made about the Control Panel manipulator, we're going to go ahead with making a ControlPanel subsystem instead of implementing this misplaced feature.

PR #5, opened 2020-02-05
Implement Color Detection

FireShredder24:
Adds a Subsystem, the ColorSubsystem, that handles the conversion of raw RGB color data from the REVRobotics ColorSensorV3 into one of the four colors on the Control Panel.


src/main/java/frc/robot/subsystems/ColorSubsystem.java
```diff
@@ -0,0 +1,97 @@
...
+/**
+ * A class to facilitate getting one of four colors from the REVRobotics
+ * ColorSensorV3
+ */
+public class ColorSubsystem extends SubsystemBase {
+  /**
+   * @param hsv The array that is used to contain the output of
+   *            java.awt.Color.RGBtoHSB()
+   */
+  public float[] hsv = new float[3];
```
terbos90803: class members should not be public.  Make this private.  Can also make final because the Java semantics of final only prevent replacing the instance of the array, they don't block altering the contents of the array.

src/main/java/frc/robot/subsystems/ColorSubsystem.java
```diff
@@ -0,0 +1,97 @@
...
+  // The ints used to store the raw ADC output of the ColorSensorV3
+  public int red, green, blue;
```
terbos90803: Same comment applies to all these members.  They should be private.

Also, these variables should be local to the method, not class members.  The only reason that hsv is a class member is to avoid reallocating the array.  The rest of these should just be locals.

src/main/java/frc/robot/subsystems/ColorSubsystem.java
```diff
@@ -0,0 +1,97 @@
...
+import edu.wpi.first.wpilibj.I2C.Port;
+import edu.wpi.first.wpilibj2.command.SubsystemBase;
+
+/**
+ * A class to facilitate getting one of four colors from the REVRobotics
+ * ColorSensorV3
+ */
+public class ColorSubsystem extends SubsystemBase {
+  /**
+   * @param hsv The array that is used to contain the output of
```
terbos90803: This is not a @param.  Remove that tag.

src/main/java/frc/robot/subsystems/ColorSubsystem.java
```diff
@@ -0,0 +1,97 @@
...
+   *            java.awt.Color.RGBtoHSB()
+   */
+  public float[] hsv = new float[3];
+  // The ints used to store the raw ADC output of the ColorSensorV3
+  public int red, green, blue;
+  // The maximum allowed absolute distance from the constant to ensure an accurate
+  // reading
+  final float tolerance = 0.025f;
```
terbos90803: This is not marked with a scope, which means it's package private.  Unless you have a really good reason for exposing constants, they should be explicitly private.

src/main/java/frc/robot/subsystems/ColorSubsystem.java
```diff
@@ -0,0 +1,97 @@
...
+  // The constants that correspond to the average hue (H) of the Control Panel
+  // colors.
+  final float kYellow = 0.25f;
```
terbos90803: To simplify the matching code, it might be better to group these constants as some form of collection.  The best collection type for this is probably a Map.  It associates keys (the number) with values (the color name).

The color names should probably be an enum set defined globally because it'll need to be shared across the APIs between subsystems and commands.

src/main/java/frc/robot/subsystems/ColorSubsystem.java
```diff
@@ -0,0 +1,97 @@
...
+  final float kCyan = 0.50f;
+  // Used to prevent repeated indexing of hsv[]
+  private float hue;
+  // A number used to store the minimum difference between the measured value and
+  // the constants
+  private float mindiff = 1f;
+  // The absolute difference between the measured value and the constants
+  private float diff;
+  // Used to store the return message
+  String color;
```
terbos90803: Why are you storing a return value as a public member?  Neither of those things should be true.  Return values should be local to their method.

src/main/java/frc/robot/subsystems/ColorSubsystem.java
```diff
@@ -0,0 +1,97 @@
...
+  // The REVRobotics ColorSensorV3
+  final ColorSensorV3 colorSensor = new ColorSensorV3(Port.kOnboard);
```
terbos90803: Private.

src/main/java/frc/robot/subsystems/ColorSubsystem.java
```diff
@@ -0,0 +1,97 @@
...
+  public void ColorSensing() {
+
+  }
+
+  public String getColor() {
+    // Sets the return string to Error so it is returned if no conditions are
+    // fulfilled
+    color = "Error";
+    // Gets the raw ADC values
+    red = colorSensor.getRed();
```
terbos90803: Don't call getRed, getGreen, and getBlue separately.  You should call getRawColor and then pick the fields out of that.  It prevents calling the sensor multiple times.

obtw, did you notice in the docs that they provide a getColor method that returns a color enum?  I suspect many teams will start with that.  The big problem with using that directly is that there are a gazillion enums and I find it unlikely that it'd land on the ones we want reliably.  We're much better off using the HSB transform.

The docs also appear to be out of sync with the code samples.  In the samples, Rev provides a color Matching class that seems to do the same thing we're doing.  I suspect they're probably using it under the hood for their getColor method.  Again, the problem I have is they're probably doing the matching in RGB space, which loses out on the simplicity and robustness of doing it in HSB space.

src/main/java/frc/robot/subsystems/ColorSubsystem.java
```diff
@@ -0,0 +1,97 @@
...
+    green = colorSensor.getGreen();
+    blue = colorSensor.getBlue();
+    // Converts to HSV
+    Color.RGBtoHSB(red, green, blue, hsv);
+    hue = hsv[0];
+    mindiff = 1f;
+    // Gets the difference between the measured value and the first constant
```
terbos90803: Obvious comments are not helpful.  This applies to many of the comments from here on.  Except in particularly obscure coding examples, the reader should be able to figure out the what pretty easily.  Make sure your comments explain the *why* of something and not the what.
Never be afraid to write a single big block comment explaining an algorithm at the top of that code instead of tagging each line.

src/main/java/frc/robot/subsystems/ColorSubsystem.java
```diff
@@ -0,0 +1,97 @@
...
+    // Gets the difference between the measured value and the second constant
+    diff = Math.abs(hue - kRed);
+    // Checks if the difference is in range
+    if (diff < mindiff) {
+      mindiff = diff;
+      color = "Red";
+    }
+    // Gets the difference between the measured value and the third constant
+    diff = Math.abs(hue - kGreen);
+    // I don't really need to tell you what this does, do I?
```
terbos90803: Really?  Snide comments are worse than no comments.

src/main/java/frc/robot/subsystems/ColorSubsystem.java
```diff
@@ -0,0 +1,97 @@
...
+    // Gets the raw ADC values
+    red = colorSensor.getRed();
+    green = colorSensor.getGreen();
+    blue = colorSensor.getBlue();
+    // Converts to HSV
+    Color.RGBtoHSB(red, green, blue, hsv);
+    hue = hsv[0];
+    mindiff = 1f;
+    // Gets the difference between the measured value and the first constant
+    diff = Math.abs(hue - kYellow);
```
terbos90803: The DRY principle applies to this section of code.  I know we wrote it this way for expediency, but it really needs to be refactored into a loop.  This will be possible when the color constants are collected in a Map that can be looped on.

terbos90803 (changes requested):

src/main/java/frc/robot/subsystems/ColorSubsystem.java
FireShredder24: I did notice their getColor and its ridiculous enum and decided exactly that.

src/main/java/frc/robot/subsystems/ColorOptions.java
```diff
@@ -0,0 +1,5 @@
+package frc.robot.subsystems;
```
terbos90803: Probably not for the subsystem package.  Is there a more central place for constants?

src/main/java/frc/robot/subsystems/ColorSubsystem.java
```diff
@@ -2,91 +2,93 @@
...
-  final float kYellow = 0.25f;
-  final float kRed = 0.10f;
-  final float kGreen = 0.35f;
-  final float kCyan = 0.50f;
+  private final float kYellow = 0.25f;
+  private final float kRed = 0.10f;
+  private final float kGreen = 0.35f;
+  private final float kCyan = 0.50f;
   // Used to prevent repeated indexing of hsv[]
   private float hue;
```
terbos90803: These variables cannot be class members, they must be locals to the method.

src/main/java/frc/robot/subsystems/ColorSubsystem.java
terbos90803: Did you need assistance with this refactor?

terbos90803 (changes requested):
Still have some cleanup and refactor to do.

src/main/java/frc/robot/subsystems/ColorSubsystem.java
FireShredder24: Yes, and also with the Map collection.  The documentation is very unhelpful.

src/main/java/frc/robot/subsystems/ColorOptions.java
FireShredder24: Yes, moved it to Utils.

src/test/java/frc/robot/subsystems/ColorSubsystemTest.java
```diff
@@ -0,0 +1,92 @@
...
+
+    ColorOptions co = cs.getColor();
+
+    assertEquals(ColorOptions.NONE, co);
+  }
+
+  @Test
+  public void testRed() {
+    RawColor rc = new RawColor(255, 153, 0, 0);
+    System.out.format("rc: Red:%d Green:%d Blue:%d\n", rc.red, rc.green, rc.blue);
```
terbos90803: unit tests should not have print statements.  I used these for getting the initial test working.  They should all come out.

src/test/java/frc/robot/subsystems/ColorSubsystemTest.java
```diff
@@ -0,0 +1,92 @@
...
+
+    ColorSubsystem cs = new ColorSubsystem(new MockColorSource(rc));
+
+    ColorOptions co = cs.getColor();
+    System.out.format("co: %s\n", co.toString());
+
+    assertEquals(ColorOptions.YELLOW, co);
+  }
+}
```
terbos90803, line 123: In addition to the checks for the 4 intended colors, there need to be a set of tests to verify that other colors are correctly rejected.

terbos90803 (changes requested):

src/test/java/frc/robot/subsystems/ColorSubsystemTest.java
FireShredder24: I picked 4 colors that are similar in RGB values to each of the 4 colors on the CP to confirm that the detection code is able to reject them.  It does.

terbos90803 (approved):
Looks good.

PR #6, opened 2020-02-21
Implement StorageSubsystem

FireShredder24:
Adds a Subsystem class that provides on-off control of the VictorSP that controls the 2 conveyance motors and the VictorSP that controls the shooter gate motor.


src/main/java/frc/robot/subsystems/StorageSubsystem.jav
```diff
@@ -0,0 +1,25 @@
+package frc.robot.subsystems;
```
jordan-powers: There’s a typo in the name of this file. It should be a ‘.java’ file, not a ‘.jav’ file.

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -93,7 +93,7 @@ public ShooterSubsystem() {
     shooterMAXRight.setIdleMode(IdleMode.kCoast);
     shooterMAXLeft.setSmartCurrentLimit(currentLimit);
     shooterMAXRight.setSmartCurrentLimit(currentLimit);
-    // Sets the left shooter motor to follow the right motor, and be inverted.
```
terbos90803: Try not to mix unrelated cleanup.

src/main/java/frc/robot/subsystems/StorageSubsystem.jav
terbos90803: upvote

src/main/java/frc/robot/subsystems/StorageSubsystem.jav
```diff
@@ -0,0 +1,25 @@
...
+public class StorageSubsystem extends SubsystemBase {
+  private final VictorSP StorageGate;
+  private final VictorSP StorageMid;
+
+  public StorageSubsystem() {
+    StorageMid = new VictorSP(Constants.STORAGE_MID_PWM_CHAN);
+    StorageGate = new VictorSP(Constants.STORAGE_GATE_PWM_CHAN);
+  }
+  public runMid() {
+    StorageMid.set(1);
```
terbos90803: No magic numbers.  Use named constants.

src/main/java/frc/robot/subsystems/StorageSubsystem.jav
```diff
@@ -0,0 +1,25 @@
...
+  }
+  public runGate() {
+    StorageGate.set(0.1);
+  }
+  public stopMotors() {
+    StorageGate.stopMotors();
```
terbos90803: You'll likely want to be able to keep the mid running while stopping the gate.  I would recommend an API more along the lines of:
public void runMid(bool run);
public void runGate(bool run);
public void stopAll();

terbos90803 (changes requested):

terbos90803:
Something else important to understand that in FRC, all motor controllers are on watchdog timers.  If you don't feed them a value on every scheduler cycle, they'll safe stop.

This means that if you want to be able to set a state once and forget about it, then you need to change your architecture to hold the desired state in the subsystem class and then set it to the motors in the subsystem's periodic method.

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
FireShredder24: Got it.  For future reference, don't fix unrelated issues you spot while working on a feature branch that are not part of said branch.  Merge conflicts are hell.

src/main/java/frc/robot/Constants.java
```diff
@@ -37,4 +37,8 @@
   // Brushed PWM addresses
   public static final int STORAGE_GATE_PWM_CHAN = 1;
   public static final int STORAGE_MID_PWM_CHAN = 2;
+
+  // Motor value constants
+  public static final double STORAGE_MID_VAL = 1;
+  public static final double STORAGE_GATE_VAL = 0.1;
```
terbos90803: Given that these values may need to be tuned, consider making them preferences instead of hardcoded.

src/main/java/frc/robot/subsystems/StorageSubsystem.java
```diff
@@ -0,0 +1,40 @@
...
+    StorageGate = new VictorSP(Constants.STORAGE_GATE_PWM_CHAN);
+  }
+
+  public void runMid(boolean run) {
+    runMid = run;
+  }
+
+  public void runGate(boolean run) {
+    runGate = run;
+  }
```
terbos90803: How will you handle reversing?

src/main/java/frc/robot/subsystems/StorageSubsystem.java
```diff
@@ -0,0 +1,40 @@
...
+    runMid = false;
+    runGate = false;
+  }
+
+  @Override
+  public void periodic() {
+    if (runMid)
+      StorageMid.set(Constants.STORAGE_MID_VAL);
+    if (runGate)
+      StorageGate.set(Constants.STORAGE_GATE_VAL);
```
terbos90803: Actively set the motor on every call.  The value changes depending on the runMid/runGate states.

terbos90803 (changes requested):

src/main/java/frc/robot/utils/MoPrefs.java
```diff
@@ -6,6 +6,7 @@
   static final double INTAKE_ROLLER_SETPOINT = 0.3;
   static final int CLIMBER_ENCODER_LIMIT = 10000;
   static final double SHOOTER_HOOD_SETPOINT = 15;
+  static final double STORAGE_SPEED_SETPOINT = 1.0;
```
FireShredder24: STORAGE_SPEED_SETPOINT seems like a slightly inappropriate name, given that setpoints always mean speed in this context.  I suggest STORAGE_SETPOINT.

FireShredder24: STORAGE_SPEED_SETPOINT seems like a slightly inappropriate name, given that setpoints always mean speed in this context. I suggest STORAGE_SETPOINT.

terbos90803 (approved):

PR #7, opened 2020-02-21
Implement IntakeSubsystem and IntakeCommand

FireShredder24:
Implements a Subsystem that controls the 2 AM-RedLine motors that spin the intake rollers and the two pneumatic pistons that actuate the intake, as well as a Command that binds said Subsystem to the controller.


src/main/java/frc/robot/Constants.java
```diff
@@ -32,7 +32,13 @@
   // SparkMAX Encoder Channels
   public static final int SHOOTER_ENCODER_CHAN_A = 1;
   public static final int SHOOTER_ENCODER_CHAN_B = 2;
-  // Pneumatic solenoid addresses
 
+  public static final int INTAKE_VICTORSP_PWM_CHAN_LF = 3;
+  public static final int INTAKE_VICTORSP_PWM_CHAN_RT = 4;
+  // Pneumatic solenoid addresses
+  public static final int INTAKE_PISTON_PCM_CHAN_LF_F = 1;
```
terbos90803: Consider renaming with higher level sense of direction.
Instead of F and R for forward and reverse, perhaps choose DEPLOY and STOW.

src/main/java/frc/robot/commands/IntakeCommand.java
```diff
@@ -0,0 +1,71 @@
...
+
+import frc.robot.subsystems.IntakeSubsystem;
+import edu.wpi.first.wpilibj2.command.CommandBase;
+import frc.robot.RobotContainer;
+
+/**
+ * An example command that uses an example subsystem.
+ */
+public class IntakeCommand extends CommandBase {
+  @SuppressWarnings({ "PMD.UnusedPrivateField", "PMD.SingularField" })
```
terbos90803: Don't suppress warnings at this time.  You need to be reminded that the code is not complete.

src/main/java/frc/robot/commands/IntakeCommand.java
```diff
@@ -0,0 +1,71 @@
...
+  private final IntakeSubsystem m_subsystem;
+  private RobotContainer m_container;
```
terbos90803: Should be final.

src/main/java/frc/robot/commands/IntakeCommand.java
```diff
@@ -0,0 +1,71 @@
...
+
+  private boolean isLowered = false;
+
+  /**
+   * Creates a new ExampleCommand.
```
terbos90803: Update comment

src/main/java/frc/robot/commands/IntakeCommand.java
```diff
@@ -0,0 +1,71 @@
...
+
+  // Called when the command is initially scheduled.
+  @Override
+  public void initialize() {
+  }
+
+  // Called every time the scheduler runs while the command is scheduled.
+  @Override
+  public void execute() {
+    if (m_container.mainController.getIntakePowerCells()) {
```
terbos90803: I know that the example code from last year did this kind of things where a command would keep running and then poll the joystick buttons.  While this does work, the intended usage of the Button commands is to trigger on a button press, update the subsystem state, and then end.
It's more reasonable for the Drive to run continuously since it's reading the analog inputs, but the buttons can be broken out separately.
We should talk about this.

src/main/java/frc/robot/commands/IntakeCommand.java
```diff
@@ -0,0 +1,71 @@
...
+
+/**
+ * An example command that uses an example subsystem.
+ */
+public class IntakeCommand extends CommandBase {
+  @SuppressWarnings({ "PMD.UnusedPrivateField", "PMD.SingularField" })
+  private final IntakeSubsystem m_subsystem;
+  private RobotContainer m_container;
+
+  private boolean isLowered = false;
```
terbos90803: You should not hold state in the Command.  State should be held in the subsystem.

src/main/java/frc/robot/controllers/DriveController.java
```diff
@@ -36,6 +36,13 @@
    */
   abstract public boolean getIntakePowerCells();
 
+  /**
+   * 
+   * @return Whether the pistons should fire that raise the intake back to its
+   *         starting position.
+   */
+  abstract public boolean getToggleIntake();
```
terbos90803: The abstract keyword is not needed for interfaces as it's implied.

src/main/java/frc/robot/subsystems/IntakeSubsystem.java
```diff
@@ -7,14 +7,42 @@
...
+      Constants.INTAKE_PISTON_PCM_CHAN_LF_R);
+  private final DoubleSolenoid intakePistonR = new DoubleSolenoid(Constants.INTAKE_PISTON_PCM_CHAN_RT_F,
+      Constants.INTAKE_PISTON_PCM_CHAN_RT_R);
+
   public IntakeSubsystem() {
+    intakeSPrt.setInverted(true);
+  }
+
+  public void runIntake() {
+    intakeSPrt.set(0.3);
```
terbos90803: 1) No magic numbers.  Use named constants.
2) Should move motor set calls into periodic method.
3) Better API is: public void runIntake(bool run);

src/main/java/frc/robot/subsystems/IntakeSubsystem.java
```diff
@@ -7,14 +7,42 @@
...
+    intakeSPlf.set(0.3);
+  }
+
+  public void stopIntake() {
+    intakeSPrt.stopMotor();
+    intakeSPlf.stopMotor();
+  }
+
+  public void raiseIntake() {
+    intakePistonL.set(DoubleSolenoid.Value.kForward);
```
terbos90803: Should replace reference to kForward etc with class constants to abstract Deploy vs Stow.

terbos90803 (changes requested):

src/main/java/frc/robot/commands/IntakeCommand.java
jordan-powers: We don't even use the PMD analyzer. I'm guessing this is left over from an example somewhere (I'm pretty sure wpilib uses PMD).

FireShredder24: Yeah, that was from the ExampleCommand.

src/main/java/frc/robot/Constants.java
```diff
@@ -36,9 +36,9 @@
...
   public static final int INTAKE_VICTORSP_PWM_CHAN_RT = 4;
   // Pneumatic solenoid addresses
-  public static final int INTAKE_PISTON_PCM_CHAN_LF_F = 1;
-  public static final int INTAKE_PISTON_PCM_CHAN_LF_R = 2;
-  public static final int INTAKE_PISTON_PCM_CHAN_RT_F = 3;
-  public static final int INTAKE_PISTON_PCM_CHAN_RT_R = 4;
+  public static final int INTAKE_PISTON_PCM_CHAN_LF_DEPLOY = 1;
+  public static final int INTAKE_PISTON_PCM_CHAN_LF_STOW = 2;
+  public static final int INTAKE_PISTON_PCM_CHAN_RT_DEPLOY = 3;
+  public static final int INTAKE_PISTON_PCM_CHAN_RT_STOW = 4;
```
terbos90803: PCM channels are numbered 0-7.

src/main/java/frc/robot/commands/intake/IntakePistonToggle.java
```diff
@@ -5,33 +5,25 @@
...
 import edu.wpi.first.wpilibj2.command.CommandBase;
-import frc.robot.RobotContainer;
 
 /**
- * An example command that uses an example subsystem.
+ * A command that uses the IntakeSubsystem.
  */
-public class IntakeCommand extends CommandBase {
-  @SuppressWarnings({ "PMD.UnusedPrivateField", "PMD.SingularField" })
+public class IntakePistonToggle extends CommandBase {
```
terbos90803: Blanket comment to all Commands:  The name of the class should have Command or at least Cmd in it.  This promotes readability of the calling code.  Whichever you choose, be consistent across all instances.

src/main/java/frc/robot/subsystems/IntakeSubsystem.java
```diff
@@ -16,37 +16,59 @@
...
-    intakePistonR.set(DoubleSolenoid.Value.kReverse);
+    intakePistonL.set(stow);
+    intakePistonR.set(stow);
+    isLowered = true;
   }
 
   @Override
   public void periodic() {
     // This method will be called once per scheduler run
+    if (runIntake && !reverseIntake) {
```
terbos90803: Calculate motor setpoint first, then always set the motor value, even if it's zero.
If you don't set the motor, you get a bunch of watchdog warnings.

terbos90803 (changes requested):

src/main/java/frc/robot/subsystems/IntakeSubsystem.java
```diff
@@ -24,6 +24,7 @@
   public boolean isLowered = false;
   private boolean runIntake = false;
   private boolean reverseIntake = false;
+  private final double rollerSetpoint = 0.3;
```
terbos90803: Should be a MoPref so we can tune it from the ShuffleBoard.

src/main/java/frc/robot/subsystems/IntakeSubsystem.java
```diff
@@ -64,11 +65,14 @@ public void lowerIntake() {
...
+      intakeSPrt.set(rollerSetpoint);
+      intakeSPlf.set(rollerSetpoint);
     } else if (runIntake && reverseIntake) {
-      intakeSPlf.set(-0.3);
-      intakeSPrt.set(-0.3);
+      intakeSPlf.set(-rollerSetpoint);
+      intakeSPrt.set(-rollerSetpoint);
+    } else {
+      intakeSPlf.set(0);
+      intakeSPrt.set(0);
```
terbos90803: Good coding practice is to only set the motor values in one place.  Do all the math/selection first, then set the values once.

Example:
```
double setpoint = runIntake ? rollerSetpoint : 0;
if (reverseIntake)
  setpoint = -setpoint;

intakeSPlf.set(setpoint);
intakeSPrt.set(setpoint);
```

terbos90803 (changes requested):

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -25,13 +35,24 @@
...
+  private XboxController xbox = new XboxController(0);
+  private LogitechF310 f310 = new LogitechF310(2);
 
-  public final DriveCommand driveCommand = new DriveCommand(driveSubsystem);
+  private final ControllerBase mainController = new ControllerBase(xbox, f310);
 
-  public final ControllerBase mainController = new ControllerBase();
+  private final JoystickButton intakeRollerFwdButton = new JoystickButton(f310, 4/* LeftBumper */);
+  private final JoystickButton intakeRollerFwdRevToggle = new JoystickButton(f310, 0/* X */);
+  private final JoystickButton intakePistonToggle = new JoystickButton(f310, 2/* B */);
```
terbos90803, line 48: There should be enums for the named buttons instead of numbers.

src/main/java/frc/robot/commands/intake/IntakePistonCmd.java
```diff
@@ -0,0 +1,53 @@
+/*----------------------------------------------------------------------------*/
```
terbos90803: Each of these command files is 99% boilerplate and 1 line of useful code.  These should all be converted to lambdas in configureButtonBindings.

src/main/java/frc/robot/subsystems/IntakeSubsystem.java
```diff
@@ -7,18 +7,72 @@
...
+import edu.wpi.first.wpilibj.VictorSP;
 import edu.wpi.first.wpilibj2.command.SubsystemBase;
+import edu.wpi.first.wpilibj.DoubleSolenoid;
+import frc.robot.Constants;
 
 public class IntakeSubsystem extends SubsystemBase {
-  /**
-   * Creates a new ExampleSubsystem.
-   */
+  private final VictorSP intakeSPrt = new VictorSP(Constants.INTAKE_VICTORSP_PWM_CHAN_LF);
```
terbos90803: Robot System says there is only one PWM mapped for both motors, so only need one Victor class.

terbos90803 (changes requested):

terbos90803 (approved):
Ok.  I think that's close enough to go.

PR #8, opened 2020-02-25
Implement ShooterHoodSubsystem

FireShredder24:
Provides a Subsystem for controlling the NEO-550 Brushless Motor that actuates the rack gear which controls the angle of deflection of Power Cells as they exit the shooter.
This Subsystem provides two types of control:
 - Fine, SmartMotion-enabled control intended for use in conjunction with vision
 - Binary, deploy-or-stow control intended for teleop.

### DIFF
--- a/src/main/java/frc/robot/Constants.java
+++ b/src/main/java/frc/robot/Constants.java
@@ -28,7 +28,6 @@ public final class Constants {
   // SparkMAX CAN addresses
   public static final int SPARKMAX_SHOOTER_CAN_ADDR_RIGHT = 5;
   public static final int SPARKMAX_SHOOTER_CAN_ADDR_LEFT = 6;
-  public static final int SPARKMAX_LOADER_CAN_ADDR = 7;
   public static final int SPARKMAX_SHOOTER_HOOD_CAN_ADDR = 3;
   // SparkMAX Encoder Channels
 
@@ -38,7 +37,7 @@ public final class Constants {
   public static final int INTAKE_PISTON_PCM_CHAN_RT_DEPLOY = 2;
   public static final int INTAKE_PISTON_PCM_CHAN_RT_STOW = 3;
   // Brushed PWM addresses
-   public static final int INTAKE_VICTORSP_PWM_CHAN = 0;
+  public static final int INTAKE_VICTORSP_PWM_CHAN = 0;
 
   // LEDs
   public static final int ADDRESSABLE_LED_PWM_ADDRESS = 5;
--- a/src/main/java/frc/robot/Constants.java
+++ b/src/main/java/frc/robot/Constants.java
@@ -29,6 +29,7 @@ public final class Constants {
   public static final int SPARKMAX_SHOOTER_CAN_ADDR_RIGHT = 5;
   public static final int SPARKMAX_SHOOTER_CAN_ADDR_LEFT = 6;
   public static final int SPARKMAX_LOADER_CAN_ADDR = 7;
+  public static final int SPARKMAX_SHOOTER_HOOD_CAN_ADDR = 8;
   // SparkMAX Encoder Channels
   public static final int SHOOTER_ENCODER_CHAN_A = 1;
   public static final int SHOOTER_ENCODER_CHAN_B = 2;
--- a/src/main/java/frc/robot/Constants.java
+++ b/src/main/java/frc/robot/Constants.java
@@ -29,10 +29,9 @@ public final class Constants {
   public static final int SPARKMAX_SHOOTER_CAN_ADDR_RIGHT = 5;
   public static final int SPARKMAX_SHOOTER_CAN_ADDR_LEFT = 6;
   public static final int SPARKMAX_LOADER_CAN_ADDR = 7;
-  public static final int SPARKMAX_SHOOTER_HOOD_CAN_ADDR = 8;
+  public static final int SPARKMAX_SHOOTER_HOOD_CAN_ADDR = 3;
   // SparkMAX Encoder Channels
-  public static final int SHOOTER_ENCODER_CHAN_A = 1;
-  public static final int SHOOTER_ENCODER_CHAN_B = 2;
+
   // Pneumatic solenoid addresses
   public static final int INTAKE_PISTON_PCM_CHAN_LF_DEPLOY = 0;
   public static final int INTAKE_PISTON_PCM_CHAN_LF_STOW = 1;
--- a/src/main/java/frc/robot/subsystems/ShooterHoodSubsystem.java
+++ b/src/main/java/frc/robot/subsystems/ShooterHoodSubsystem.java
@@ -1,11 +1,14 @@
 package frc.robot.subsystems;
 
-import com.REVRobotics.CANSparkMAX;
+import com.REVRobotics.CANSparkMax;
 import edu.wpi.first.wpilibj2.SubsystemBase;
 
 public class ShooterHoodSubsystem extends SubsystemBase {
-  
+  private final CANSparkMax hoodNEO;
   public ShooterHoodSubsystem() {
-  
+    
+  }
+  public void moveHood(double moveRequest) {
+     
   }
 }
--- a/src/main/java/frc/robot/subsystems/ShooterHoodSubsystem.java
+++ b/src/main/java/frc/robot/subsystems/ShooterHoodSubsystem.java
@@ -1,0 +1,11 @@
+package frc.robot.subsystems;
+
+import com.REVRobotics.CANSparkMAX;
+import edu.wpi.first.wpilibj2.SubsystemBase;
+
+public class ShooterHoodSubsystem extends SubsystemBase {
+  
+  public ShooterHoodSubsystem() {
+  
+  }
+}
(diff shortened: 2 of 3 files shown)


src/main/java/frc/robot/Constants.java
```diff
@@ -29,6 +29,7 @@
   public static final int SPARKMAX_SHOOTER_CAN_ADDR_RIGHT = 5;
   public static final int SPARKMAX_SHOOTER_CAN_ADDR_LEFT = 6;
   public static final int SPARKMAX_LOADER_CAN_ADDR = 7;
+  public static final int SPARKMAX_SHOOTER_HOOD_CAN_ADDR = 8;
```
terbos90803: The CAN addresses have not been selected for any of the SparkMAXs.  Sam is supposed to do this and up date the system chart at tonight's meeting.  Bug him until he gets this done.

src/main/java/frc/robot/Constants.java
```diff
@@ -29,6 +29,7 @@
...
   // SparkMAX Encoder Channels
   public static final int SHOOTER_ENCODER_CHAN_A = 1;
   public static final int SHOOTER_ENCODER_CHAN_B = 2;
```
terbos90803: I'm not sure what encoder channels are.  I don't think it's what you want.  We should sort this out.

src/main/java/frc/robot/subsystems/ShooterHoodSubsystem.java
```diff
@@ -0,0 +1,107 @@
...
+  private CANPIDController hoodPID;
+
+  private final double kP = 5e-5;
+  private final double kI = 1e-6;
+  private final double kD = 0;
+  private final double kIz = 0;
+  private final double kFF = 0.000156;
+  private final double kMaxOutput = 0.3;
+  private final double kMinOutput = -0.3;
+  private final double allowedErr = 0;
```
terbos90803, line 31: These are the constants for the full NEO.  I've not seen any published constants for the NEO550.  So we can start with these, but we'll likely have to tune them.

src/main/java/frc/robot/subsystems/ShooterHoodSubsystem.java
```diff
@@ -0,0 +1,107 @@
...
+
+  private double hoodPos;
+  private boolean reliableZero;
+  private boolean enableLimit;
+  public boolean isDeployed = false;
+
+  private double minVel = 0;
+  private double maxVel = 0;
+  private double maxAcc = 0;
```
terbos90803: These will need to be non-zero.

src/main/java/frc/robot/subsystems/ShooterHoodSubsystem.java
```diff
@@ -0,0 +1,107 @@
...
+    hoodLimitSwitch.enableLimitSwitch(true);
+
+    hoodPID.setSmartMotionAccelStrategy(CANPIDController.AccelStrategy.kTrapezoidal, 0);
+    hoodPID.setSmartMotionMaxVelocity(maxVel, 0);
+    hoodPID.setSmartMotionMaxAccel(maxAcc, 0);
+    hoodPID.setSmartMotionMinOutputVelocity(minVel, 0);
+    hoodPID.setSmartMotionAllowedClosedLoopError(allowedErr, 0);
+
+    hoodNEO.getForwardLimitSwitch(CANDigitalInput.LimitSwitchPolarity.kNormallyOpen).enableLimitSwitch(false);
+    hoodLimitSwitch.enableLimitSwitch(enableLimit);
```
terbos90803: Redundant with line 49.

src/main/java/frc/robot/subsystems/ShooterHoodSubsystem.java
```diff
@@ -0,0 +1,107 @@
...
+
+    hoodNEO.setIdleMode(IdleMode.kCoast);
```
terbos90803: Should be kBrake.

src/main/java/frc/robot/subsystems/ShooterHoodSubsystem.java
```diff
@@ -0,0 +1,107 @@
...
+  private final double kD = 0;
+  private final double kIz = 0;
+  private final double kFF = 0.000156;
+  private final double kMaxOutput = 0.3;
+  private final double kMinOutput = -0.3;
+  private final double allowedErr = 0;
+
+  private double hoodPos;
+  private boolean reliableZero;
+  private boolean enableLimit;
```
terbos90803: I'm not sure if we'd ever want to disable the limit after initial calibration.

src/main/java/frc/robot/subsystems/ShooterHoodSubsystem.java
```diff
@@ -0,0 +1,107 @@
...
+    hoodNEO.getForwardLimitSwitch(CANDigitalInput.LimitSwitchPolarity.kNormallyOpen).enableLimitSwitch(false);
+    hoodLimitSwitch.enableLimitSwitch(enableLimit);
+
+    hoodNEO.setIdleMode(IdleMode.kCoast);
+  }
+
+  public void moveHood(double posRequest) {
+    // Used for autonomous and vision-tied control of the shooter hood.
+    hoodPID.setReference(posRequest, ControlType.kSmartMotion, 0);
+    enableLimit = false;
```
terbos90803: don't disable limit.

src/main/java/frc/robot/subsystems/ShooterHoodSubsystem.java
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  public void deployHood() {
+    while (hoodPos < maxHoodPos) {
+      hoodPID.setReference(-0.1, ControlType.kSmartMotion, 0);
```
terbos90803: Never put while loops in code like this.  Blocking code like this will infinite loop and kill the robot.

The correct procedure is to simply set the state and return.

Keep in mind that setting the Reference is the same as setting any motor control.  You have to do it every cycle.  So again, this is the type of thing to save state here and then apply the state in the periodic method.

src/main/java/frc/robot/subsystems/ShooterHoodSubsystem.java
```diff
@@ -0,0 +1,107 @@
...
+    }
+    isDeployed = true;
+  }
+
+  public void stowHood() {
+    while (!hoodLimitSwitch.get()) {
```
terbos90803: Same as above.

terbos90803 (changes requested):

src/main/java/frc/robot/Constants.java
```diff
@@ -29,10 +29,9 @@
   public static final int SPARKMAX_SHOOTER_CAN_ADDR_RIGHT = 5;
   public static final int SPARKMAX_SHOOTER_CAN_ADDR_LEFT = 6;
   public static final int SPARKMAX_LOADER_CAN_ADDR = 7;
```
terbos90803: What is "Loader"

src/main/java/frc/robot/subsystems/ShooterHoodSubsystem.java
```diff
@@ -55,29 +55,23 @@ public ShooterHoodSubsystem() {
...
-    enableLimit = false;
   }
 
   public void deployHood() {
-    while (hoodPos < maxHoodPos) {
-      hoodPID.setReference(-0.1, ControlType.kSmartMotion, 0);
-    }
-    isDeployed = true;
+    deploy = true;
+    stow = false;
```
terbos90803: If deploy and stow are always opposites (which they are), they should never be represented by two different variables.

Options are:

- Single boolean called either stow OR deploy
- Enum indicating different state options, of which there are currently 2.

src/main/java/frc/robot/subsystems/ShooterHoodSubsystem.java
```diff
@@ -94,11 +88,17 @@ public void periodic() {
     if (hoodLimitSwitch.get()) {
       zeroHood();
     }
+    double hoodSetpoint = 0;
+    if (deploy)
+      hoodSetpoint = MoPrefs.getShooterHoodSetpoint();
+    if (stow)
+      hoodSetpoint = 0;
```
terbos90803: Cleanup for single controlling boolean.

terbos90803 (changes requested):

src/main/java/frc/robot/Constants.java
FireShredder24: I think I put that there while working on the shooter.  It's a deprecated version of the shooter gate, which is under StorageSubsystem

src/main/java/frc/robot/subsystems/ShooterHoodSubsystem.java
```diff
@@ -91,7 +88,7 @@ public void periodic() {
     double hoodSetpoint = 0;
     if (deploy)
       hoodSetpoint = MoPrefs.getShooterHoodSetpoint();
-    if (stow)
+    else
```
terbos90803: You've already initialized hoodSetpoint to 0 on 88.  No need to set it again in the else.

terbos90803 (changes requested):

terbos90803 (approved):

PR #9, opened 2020-02-26
Impl DriveCommand

FireShredder24:
Fills out the DriveCommand, which passes joystick inputs to the DriveSubsystem to effectuate movement of the robot.


src/main/java/frc/robot/commands/DriveCommand.java
```diff
@@ -9,13 +9,16 @@
 
 import frc.robot.subsystems.DriveSubsystem;
 import edu.wpi.first.wpilibj2.command.CommandBase;
+import frc.robot.RobotContainer;
 
 public class DriveCommand extends CommandBase {
   @SuppressWarnings({ "PMD.UnusedPrivateField", "PMD.SingularField" })
```
jordan-powers: As before, remove the extraneous SupressWarnings

src/main/java/frc/robot/commands/DriveCommand.java
```diff
@@ -28,6 +31,9 @@ public void initialize() {
   // Called every time the scheduler runs while the command is scheduled.
   @Override
   public void execute() {
+    drive_subsystem.drive(
+        Math.min(m_container.mainController.getMoveRequest(), m_container.mainController.getSpeedLimiter()),
```
jordan-powers: I'm pretty sure you want to be using a map here, not a min.

Although, it's probably architecturally better to just deal with all the input processing (speed limiting, deadzone mapping, input curving) in the DriveController class and just pass the desired move request straight to the drive subsystem.
So you probably don't need the `getSpeedLimiter` method in the ControllerBase.

I figure you're probably using it because we used it in previous years. However, I'm pretty sure that's a holdover from when we were doing Iterative robots and not Command-Based robots. To me it seems better to separate the input processing logic from the driving logic.

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -29,7 +29,7 @@
   public final ShooterSubsystem shooterSubsystem = new ShooterSubsystem();
   public final IntakeSubsystem IntakeSubsystem = new IntakeSubsystem();
 
-  public final DriveCommand driveCommand = new DriveCommand(driveSubsystem);
+  public final DriveCommand driveCommand = new DriveCommand(driveSubsystem, this);
 
   public final ControllerBase mainController = new ControllerBase();
```
jordan-powers: So we're not using a SendableChooser to put the control map selection on the SmartDashboard?

src/main/java/frc/robot/commands/DriveCommand.java
jordan-powers: In fact, now that I look at the code in ControllerBase, you have MOVE_CURVE and TURN_CURVE constants (which should probably be private), but you never use them.

src/main/java/frc/robot/RobotContainer.java
terbos90803: The switch to the intended usage of button triggers makes variable control mappings a lot harder.  It was my recommendation to go with the Button command objects at this time and we'll worry about variable mappings later.

src/main/java/frc/robot/commands/DriveCommand.java
```diff
@@ -9,13 +9,16 @@
...
 import edu.wpi.first.wpilibj2.command.CommandBase;
+import frc.robot.RobotContainer;
 
 public class DriveCommand extends CommandBase {
   @SuppressWarnings({ "PMD.UnusedPrivateField", "PMD.SingularField" })
   private final DriveSubsystem drive_subsystem;
+  private final RobotContainer m_container;
 
-  public DriveCommand(DriveSubsystem subsystem) {
+  public DriveCommand(DriveSubsystem subsystem, RobotContainer container) {
```
terbos90803: Would be better to only pass in the controller instead of the whole container.

terbos90803 (changes requested):

src/main/java/frc/robot/commands/DriveCommand.java
FireShredder24: May I have some clarification on the first part(specifically, dealing with all input processing in DriveController)?  I'm a little confused.  Wouldn't the Command be the best place to do that?

jordan-powers: Well, I may be mistaken, but to me it makes the most sense that the DriveController determines what speed the user wants to move at, and the DriveCommand tells to DriveSubsystem to drive at that speed.
The DriveController does some processing (dead zones, input curves, speed limiting) to better interpret what the user wants, but interpreting the user’s desired speed is the DriveController’s job.

This is especially relevant for different types of controller. What processing you do for an Xbox controller is likely different to the processing you do for a Joystick. However, you want one single DriveCommand to be in charge of driving, no matter where the user input is coming from.

terbos90803: I concur with Jordan that there should be a logical separation of responsibility between the layers.  There are some available options for slicing within the stack, but the ordering is pretty fixed.
Fixed order of operations:

1. read raw user inputs
1. condition inputs (deadzone, curves, etc) to make values linear and normalized (-1,0,1)
1. apply additional constraints/conditions like speed scaling
1. apply closed-loop, fly-by-wire corrections (auto stability, etc) at the move/turn level
1. translate move/turn requests into drive-specific requests (move/turn -> left/right)
1. apply additional conditioning to drive-specific requests (motor velocity PID)

Clearly, 1-2 belong in the DriveController, and 5-6 belong in the DriveSubsystem (6 actually lives in the Falcons this year).  I would argue that 4 would typically end up in the DriveSubsystem, but it could be refactored into a separate layer if it has different interactions between auton and teleop modes.  Note that none of these steps has a specific affinity to DriveCommand.

So the question comes down to where to put 3.  The extreme ends of the spectrum are:

- Bake it into any one of the layers: Controller, Command, Subsystem, based on where you *think* you want to restrict its scope of influence.  i.e. If you're sure you only ever want it to affect a particular Controller, then put it there.  If you're sure you want to apply to every possible mode of control, then put it in the subsystem.
- Create a ConditionedDriveSubsystem that manages the inputs from the DriveCommand and then passes them along to the lower-level DriveSubsystem.

Breaking it out as an independent layer gives you the most flexibility, but also the most overhead.  Given the fact that we'll likely need flexibility vs performance, I would recommend going the ConditionedDriveSubsystem route.  This allows you to create button mappings and commands that can adjust the speed scaler and then the speed scaler conditions the inputs to the DriveSubsystem.

Bottom line is that it appears we're keen to keep the Commands as lightweight as possible.  This increases our ability to use InstantCommands: https://docs.wpilib.org/en/latest/docs/software/commandbased/convenience-features.html

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -90,7 +94,7 @@ private void configureButtonBindings() {
    *
    * @return the command to run in autonomous
    */
-  public Command getAutonomousCommand() {
+  public ParallelCommandGroup getAutonomousCommand() {
```
terbos90803: Why did you change this return type?  It should still be Command.

src/main/java/frc/robot/Robot.java
```diff
@@ -20,7 +21,8 @@
  */
 public class Robot extends TimedRobot {
 
-  private Command m_autonomousCommand;
+  private ParallelCommandGroup autonCommandGroup;
```
terbos90803: At this level, all commands should be Command type.  It is not correct for the Robot class to know what specific type of command is being scheduled.

src/main/java/frc/robot/Robot.java
```diff
@@ -95,9 +97,11 @@ public void teleopInit() {
     // teleop starts running. If you want the autonomous to
     // continue until interrupted by another command, remove
     // this line or comment it out.
-    if (m_autonomousCommand != null) {
-      m_autonomousCommand.cancel();
+    if (autonCommandGroup != null) {
+      autonCommandGroup.cancel();
     }
+    teleopDriveCommand.schedule(false);
```
terbos90803: Where did teleopDriveCommand come from?  It is never initialized here (and shouldn't be).  It should come from RobotContainer in a similar manner to the autonomous command.

src/main/java/frc/robot/subsystems/ConditionedDriveSubsystem.java
```diff
@@ -0,0 +1,38 @@
+package frc.robot.subsystems;
+
+import edu.wpi.first.wpilibj2.command.SubsystemBase;
+import org.usfirst.frc.team4999.utils.Utils;
+
+public class ConditionedDriveSubsystem extends SubsystemBase {
```
terbos90803: To enable flexibility in changing the Drive stack, both ConditionedDriveSubsystem and the real DriveSubsystem should implement the same interface.  This allows the DriveCommand to not know which one it's actually talking to.  I can help you with this if it isn't obvious what to do.

terbos90803 (changes requested):

src/main/java/frc/robot/subsystems/ConditionedDriveSubsystem.java
```diff
@@ -0,0 +1,38 @@
...
+    spd_limit_index = spd_limits.length - 1;
+    m_subsystem = subsystem;
+  }
+
+  public void setSpeedLimit(double limit) {
+    spd_limit = limit;
+  }
+
+  public void mapToLimit(double moveRequest, double turnRequest) {
+    m_subsystem.drive(Utils.map(moveRequest, -1, 1, -spd_limit, spd_limit), turnRequest);
```
jordan-powers: Not sure, but you might want to speed-limit the turn request too

src/main/java/frc/robot/subsystems/ConditionedDriveSubsystem.java
```diff
@@ -0,0 +1,38 @@
...
+    ++spd_limit_index;
+    if (spd_limit_index >= spd_limits.length)
+      spd_limit_index = spd_limits.length - 1;
+    spd_limit = spd_limits[spd_limit_index];
+  }
+
+  public void decSpeedLimit() {
+    --spd_limit_index;
+    if (spd_limit_index > 0)
+      spd_limit_index = 0;
```
jordan-powers: Hmm, some faulty reasoning here.
You only want to set the `spd_limit_index` to zero if its *less than* zero.
The way you have it now, whenever you call `decSpeedLimit()`, the `spd_limit_index` is going to be set to zero, no matter what it is.

src/main/java/frc/robot/subsystems/ConditionedDriveSubsystem.java
jordan-powers: Should this be a Subsystem? Are we using any of the features of SubsystemBase?
Might be better to have it as a plain Java class.

src/main/java/frc/robot/commands/DriveCommand.java
```diff
@@ -0,0 +1,54 @@
...
+  // Called when the command is initially scheduled.
+  @Override
+  public void initialize() {
+    stop();
+  }
+
+  // Called every time the scheduler runs while the command is scheduled.
+  @Override
+  public void execute() {
+    m_cdss.mapToLimit(m_controller.getMoveRequest(), m_controller.getTurnRequest());
```
jordan-powers: Maybe instead of having the conditioner hold a reference to the DriveSubsystem (which I feel somewhat breaks the idea of the command-subsystem structure), you could get the conditioned value from the conditioner then feed it to the DriveSubsystem?
```
double conditioned_mr = conditioner.conditionMoveRequest(m_controller.getMoveRequest());
double conditioned_tr = conditioner.conditionTurnRequest(m_controller.getTurnRequest());
m_drive_subsystem.drive(conditioned_mr, conditioned_tr);
```

src/main/java/frc/robot/subsystems/ConditionedDriveSubsystem.java
jordan-powers: Or maybe break out a separate `DriveConditioner` interface, then have the `DriveCommand` feed values to the `DriveConditioner`, get the conditioned values, and feed the conditioned values to the `DriveSubsystem`?

src/main/java/frc/robot/subsystems/ConditionedDriveSubsystem.java
jordan-powers: A cleaner solution might be something like
`spd_limit_index = Math.max(spd_limit_index - 1, 0);`

jordan-powers (changes requested):
I agree with @terbos90803 that this should probably be restructured.
Also the bug in `decSpeedLimit()` is pretty critical, as it could severely limit our competitiveness if we're always stuck at 25% power.

src/main/java/frc/robot/subsystems/ConditionedDriveSubsystem.java
```diff
@@ -0,0 +1,39 @@
+package frc.robot.subsystems;
+
+import org.usfirst.frc.team4999.utils.Utils;
+
+public class ConditionedDriveSubsystem extends DriveSubsystem {
```
jordan-powers: It still doesn't sit right with me that you're using a chain of subsystems for the input processing.
- The docs describe subsystems as "an abstraction for a collection of robot hardware." Using a subsystem for pure data processing that then passes that data to a different subsystem feels like the kind of command misuse we're trying to avoid this year.
- As far as I can tell, calling `require()` with an instance of ConditionedDriveSubsystem doesn't call `require()` with the FalconDriveSubsystem held by ConditionedDriveSubsystem. I guess that wouldn't be too big of an issue if the only reference to the FalconDriveSubsystem was in the ConditionedDriveSubsystem, but the fact that it's passed in as a constructor parameter worries me. I feel there's a great potential that one command could `require()` the ConditionedDriveSubsystem and the other command could `require()` the FalconDriveSubsystem. Since they're two different subsystems the two commands would be allowed to run in parallel, but they would be fighting for the same motors. This could cause unexpected, potentially dangerous behavior.
- The ConditionedDriveSubsystem doesn't use any of the features of SubsystemBase, as far as I can tell. It seems like a waste to give it all the resources and scheduling overhead of a full-blown subsystem, then not use any of it.

src/main/java/frc/robot/subsystems/ConditionedDriveSubsystem.java
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  public void drive(double moveRequest, double turnRequest) {
+    m_subsystem.drive(Utils.map(moveRequest, -1, 1, -spd_limit, spd_limit), turnRequest);
+  }
+
+  public void incSpeedLimit() {
+    ++spd_limit_index;
+    if (spd_limit_index >= spd_limits.length)
+      spd_limit_index = spd_limits.length - 1;
```
jordan-powers: I was hoping you'd notice that a similar simplification could be made here as well.
I guess it's not necessary, but it does help improve readability.
`spd_limit_index = Math.min(spd_limit_index + 1, spd_limits.length - 1);`

jordan-powers (changes requested):
I really don't like the chaining of Subsystems.
Perhaps you could explain why you're doing it that way?

src/main/java/frc/robot/subsystems/ConditionedDriveSubsystem.java
FireShredder24: > @terbos90803 Breaking it out as an independent layer gives you the most flexibility, but also the most overhead. Given the fact that we'll likely need flexibility vs performance, I would recommend going the ConditionedDriveSubsystem route. This allows you to create button mappings and commands that can adjust the speed scaler and then the speed scaler conditions the inputs to the DriveSubsystem.

This is what Terry said before that prompted me to go down this route.  What you see here is the result of a lot of discussion in real life today.  You can ask Terry if you want reasons for pursuing this path.

The decision to make ConditionedDriveSubsystem a Subsystem at all was made because Terry wanted to fulfill the "high cohesion, low coupling" rule by making CDSS and FDSS both impls of an abstract class, DSS.  The advantages of this are well known to you, I'm sure, although I strongly agree that making CDSS a Subsystem is a waste of resources.

To clarify, the current state of the branch is its intended usage.  FalconDriveSS is not intended to be used directly by any Commands that run during teleop, only take input from ConditionedDSS.

terbos90803: Ok.  So we all agree on the outline of operations in the drive stack.  The only detail to be resolved is the slicing and packaging.  While using a subsystem for the drive conditioning does work, it's not the most elegant.

I'm going to make this proposal:

- The controller layer remains as-is.  It normalizes button/axis events but does not hold any state.
- The control conditioning state is held in a separate class that is accessible by the control events and the drive command.
- The command does not directly hold state, but is aware of three things:
1. The controller
1. The conditioner
1. The drive subsystem
- The command coordinates usage of those three elements by getting requests, running them through the conditioner, then sending them to the subsystem.

While this will work, it places the onus of stack composition within the command impl.  Again, this is a case of it works, but it seems like it could be better.

At this point, let's just go with this and sort it out later.  We're out of time for polishing at this level.

src/main/java/frc/robot/subsystems/ConditionedDriveSubsystem.java
```diff
@@ -0,0 +1,39 @@
...
+    spd_limit = spd_limits[spd_limit_index];
+  }
+
+  public void decSpeedLimit() {
+    spd_limit_index = Math.max(--spd_limit_index, 0);
```
terbos90803: Using -- in this context is wasteful since it will get overwritten by the assignment anyway.  It's more clear to simply use spd_limit_index - 1

src/main/java/frc/robot/subsystems/FalconDriveSubsystem.java
```diff
@@ -0,0 +1,173 @@
...
+  }
+
+  public void resetEncoders() {
+    leftEnc.setQuadraturePosition(0, 0); // Resets both encoders so they count from zero.
+    rightEnc.setQuadraturePosition(0, 0);
+  }
+
+  /**
+   * Averages the speed of both drive sides to get mean forward velocity
+   */
```
terbos90803: Comment without a method.

src/main/java/frc/robot/subsystems/FalconDriveSubsystem.java
```diff
@@ -0,0 +1,173 @@
...
+
+  public void stop() {
+    leftFront.stopMotor();
+    rightFront.stopMotor();
+  }
```
terbos90803, line 161: Method without a comment.

terbos90803 (changes requested):

terbos90803 (approved):
Whew

PR #11, opened 2020-03-02
Implement the Climber

terbos90803:
Add the climber subsystem and button bindings.


FireShredder24 (commented):
As far as I can tell, this is good.  It's obviously meant to be used with a non-instant Command, which will require some discussion, but otherwise it looks good.

src/main/java/frc/robot/subsystems/ClimberSubsystem.java
```diff
@@ -0,0 +1,54 @@
...
+
+  public void stow() {
+    if (limit.get())
+      stop();
+    else
+      climberSP.set(CLIMB_STOW);
+  }
+
+  public void climb() {
+    if (encoder.get() > MoPrefs.getClimberEncoderLimit())
```
jordan-powers: So we're assuming the climbing encoder will always have a valid zero before we start climbing?

jordan-powers (changes requested):

FireShredder24 (commented):
Looks good.

FireShredder24 (dismissed):
Ready to merge as far as I can tell.

FireShredder24 (approved):
Looks good to me.

jordan-powers (approved):
Looks good

PR #15, opened 2020-03-05
Simulation fixes

jordan-powers:
Okay, I admit the DriveConditioning isn't _technically_ a bugfix. However, I only noticed that the curves weren't being added because I was simulating the code and watching the motor outputs.

### DIFF
--- a/src/main/java/frc/robot/subsystems/conditioners/ComposedConditioner.java
+++ b/src/main/java/frc/robot/subsystems/conditioners/ComposedConditioner.java
@@ -11,8 +11,8 @@ public class ComposedConditioner implements DriveConditioner {
   @Override
   public double conditionMove(double moveRequest) {
     double value = moveRequest;
-    for (int i = 0; i < conditioners.length; i++) {
-      value = conditioners[i].conditionMove(value);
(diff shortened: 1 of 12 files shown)


src/main/java/frc/robot/subsystems/conditioners/ComposedConditioner.java
```diff
@@ -0,0 +1,29 @@
...
+  private final DriveConditioner[] conditioners;
+
+  public ComposedConditioner(DriveConditioner... conditioners) {
+    this.conditioners = conditioners;
+  }
+
+  @Override
+  public double conditionMove(double moveRequest) {
+    double value = moveRequest;
+    for (int i = 0; i < conditioners.length; i++) {
```
terbos90803: This could be written a bit more concisely as:
for (DriveConditioner cond : conditioners)
  value = cond.conditionMove(value);

terbos90803 (commented):
Thanks for jumping in and doing all this.  I see you're working 3rd shift these days.

terbos90803 (approved):
Fixed it.

PR #22, opened 2020-03-10
Competition Changes

jordan-powers:
Judging by my previous experiences with rushed competition changes, I assume this probably needs some rework.


src/main/java/frc/robot/RobotContainer.java
```diff
@@ -44,49 +53,79 @@
...
-  private final ShooterSubsystem shooterSubsystem = new ShooterSubsystem(shooterHoodSubsystem);
-  private final StorageSubsystem storageSubsystem = new StorageSubsystem(powerDistributionPanel);
-
-  private final SpeedLimitConditioner speedLimitConditioner = new SpeedLimitConditioner();
-  private final ReverseConditioner reverseConditioner = new ReverseConditioner();
+  // --------------------------------------Shuffleboard----------------------------------------
+  public static final ShuffleboardTab matchTab = Shuffleboard.getTab("Match");
+  public static final ShuffleboardTab outreachTab = Shuffleboard.getTab("Outreach");
+  public static final ShuffleboardTab testTab = Shuffleboard.getTab("Test Tune");
+  public static final ShuffleboardTab limelightTab = Shuffleboard.getTab("Limelight");
```
terbos90803: Should all be private non-static.  The composition rules are clearly moving away from static usage.

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -44,49 +53,79 @@
...
+      .alongWith(new RunCommand(shooterSubsystem::shoot, shooterSubsystem));
+
+  // private final Command shootCommand = new
+  // RunCommand(shooterHoodSubsystem::deployHood, shooterHoodSubsystem)
+  // .withInterrupt(() -> shooterHoodSubsystem.getFullyDeployed() &&
+  // shooterHoodSubsystem.hasReliableZero())
+  // .andThen(new RunCommand(shooterSubsystem::shoot, shooterSubsystem)
+  // .alongWith(new RunCommand(shooterHoodSubsystem::deployHood,
+  // shooterHoodSubsystem)));
+
```
terbos90803: cleanup

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -109,23 +150,34 @@ public RobotContainer() {
...
+    shooterShoot.whileHeld(shootCommand);
+
+    // --------------------------------------Storage------------------------------------------------
 
-    shooterShoot.whenPressed(new InstantCommand(shooterSubsystem::shoot, shooterSubsystem, shooterHoodSubsystem));
-    // Purge should also reverse storage and intake. Need to work out best way to do
-    // that.
-    purge.whenPressed(new InstantCommand(shooterSubsystem::purge, shooterSubsystem, shooterHoodSubsystem));
+    // shooterShoot.whenPressed(new InstantCommand(shooterSubsystem::shoot,
+    // shooterSubsystem, shooterHoodSubsystem));
```
terbos90803: cleanup

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -134,8 +186,7 @@ private void configureButtonBindings() {
    * @return the command to run in autonomous
    */
   public Command getAutonomousCommand() {
-    // An ExampleCommand will run in autonomous
-    return autonomousCommand;
+    return shootFromLine;
```
terbos90803: Could be better.  Hardcoding the selection is bad.  SendableChooser is best.

src/main/java/frc/robot/Constants.java
```diff
@@ -20,14 +20,14 @@
...
 
   // Falcon CAN addresses
-  public static final int FALCON_DRIVE_LEFT_FRONT_CAN_ADDR = 1;
-  public static final int FALCON_DRIVE_LEFT_REAR_CAN_ADDR = 2;
-  public static final int FALCON_DRIVE_RIGHT_FRONT_CAN_ADDR = 3;
-  public static final int FALCON_DRIVE_RIGHT_REAR_CAN_ADDR = 4;
+  public static final int FALCON_DRIVE_LEFT_FRONT_CAN_ADDR = 3;
+  public static final int FALCON_DRIVE_LEFT_REAR_CAN_ADDR = 4;
+  public static final int FALCON_DRIVE_RIGHT_FRONT_CAN_ADDR = 1;
+  public static final int FALCON_DRIVE_RIGHT_REAR_CAN_ADDR = 2;
```
jordan-powers, line 26: Please update the Robot System Chart.

src/main/java/frc/robot/commands/ShootFromLine.java
```diff
@@ -0,0 +1,52 @@
...
+    if (13 > time && time > 5) {
+      m_storage.run();
+      m_shooter.shootAuto();
+    } else if (time < 5) {
+      m_drive.drive(0.25, 0);
+      m_storage.stop();
+      m_shooter.idle();
+    } else {
+      m_drive.stop();
+    }
```
terbos90803: There are much better ways of doing this sequencing.  That was a big basis of the Command framework.  Each of these steps can be packaged as a smaller Command (e.g. as a RunCommand) and then sequenced by chaining them together.  It's much more readable and maintainable.

src/main/java/frc/robot/commands/ShootFromWall.java
```diff
@@ -0,0 +1,49 @@
...
+    if (time > 5) {
+      m_storage.run();
+    } else if (time > 1) {
+      m_storage.stop();
+    }
+    if (time > 3) {
+      m_shooter.shoot();
+    } else if (time > 1) {
+      m_shooter.idle();
+    }
```
terbos90803: Rework as sequenced commands.

src/main/java/frc/robot/controllers/ControllerBase.java
```diff
@@ -20,9 +20,20 @@ public double getMoveRequest() {
     return moveRequest;
   }
 
+  public double getShootPowerCells() {
+    return Utils.deadzone(xbox.getTriggerAxis(Hand.kRight), 0.2);
```
terbos90803: Why is this still here?  I thought we moved this to a TriggerButton

src/main/java/frc/robot/controllers/ControllerBase.java
```diff
@@ -20,9 +20,20 @@ public double getMoveRequest() {
...
     return turnRequest;
   }
 
+  public boolean getIncHoodPos() {
+    return xbox.getPOV() == 180;
+  }
+
+  public boolean getDecHoodPos() {
+    return xbox.getPOV() == 0;
+  }
```
terbos90803, line 32: Probably should be packaged differently.  Not positive about that just yet...  Argue why it should be here.

src/main/java/frc/robot/commands/AutonDriveCommand.java
```diff
@@ -27,6 +36,25 @@ public void initialize() {
   // Called every time the scheduler runs while the command is scheduled.
   @Override
   public void execute() {
+    LimelightData data = limelight.getData();
+    double turnRequest;
+    double moveRequest;
+    double distance;
+
+    if (data.valid()) {
+      turnRequest = Utils.map(data.xCoord(), -Limelight.RANGE_X, Limelight.RANGE_X, -0.25, 0.25);
```
jordan-powers: 0.25 is a magic number.

Also, I highly recommend you use the [trajectory generator](https://docs.wpilib.org/en/latest/docs/software/examples-tutorials/trajectory-tutorial/creating-following-trajectory.html#generating-the-trajectory)

src/main/java/frc/robot/commands/AutonDriveCommand.java
```diff
@@ -27,6 +36,25 @@ public void initialize() {
...
+      moveRequest = 0; // Utils.map(data.dist(), -Limelight.RANGE_Y, Limelight.RANGE_Y, -1.0, 1.0);
+      distance = data.dist();
+      met = false; // data.targetMet();
```
jordan-powers: What's the point of the `met` variable if it's always false?

src/main/java/frc/robot/subsystems/FalconDriveSubsystem.java
```diff
@@ -86,6 +86,7 @@ public FalconDriveSubsystem() {
     // These should always be opposites
     // If the robot drives backwards, flip both
     leftFront.setInverted(true);
+    leftRear.setInverted(true);
     rightFront.setInverted(false);
```
terbos90803: For completeness, rightRear should also be listed (or rightFront removed) to avoid confusion why imbalanced.

src/main/java/frc/robot/commands/DriveToWall.java
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  @Override
+  public void execute() {
+    double time = Timer.getMatchTime();
+    if (time > 10) {
+      m_subsystem.drive(0.25, 0);
+    } else {
+      m_subsystem.stop();
+    }
```
jordan-powers: I highly recommend you [follow a trajectory](https://docs.wpilib.org/en/latest/docs/software/examples-tutorials/trajectory-tutorial/trajectory-tutorial-overview.html)

src/main/java/frc/robot/commands/ShootFromLine.java
```diff
@@ -0,0 +1,52 @@
...
+  @Override
+  public void initialize() {
+
+  }
+
+  @Override
+  public void execute() {
+    double time = Timer.getMatchTime();
+    m_intake.raiseIntake();
+    if (13 > time && time > 5) {
```
jordan-powers: The `time > 5` clause is redundant

src/main/java/frc/robot/commands/ShootFromWall.java
```diff
@@ -0,0 +1,49 @@
...
+  @Override
+  public void initialize() {
+
+  }
+
+  @Override
+  public void execute() {
+    double time = Timer.getMatchTime();
+    if (time > 10) {
+      m_subsystem.drive(0.25, 0);
```
jordan-powers: I highly recommend you [follow a trajectory](https://docs.wpilib.org/en/latest/docs/software/examples-tutorials/trajectory-tutorial/trajectory-tutorial-overview.html)

src/main/java/frc/robot/subsystems/IntakeSubsystem.java
```diff
@@ -15,16 +15,18 @@
...
   private final int SAFE_COOLDOWN_MS = 1000;
-  private final double UNSAFE_CURRENT_LIMIT = 30; // amperes
+  private final double UNSAFE_CURRENT_LIMIT = 20; // amperes
   private final int UNSAFE_CURRENT_TIMEOUT_MS = 1000;
 
   private final VictorSP intakeSP;
+  private final VictorSP intakeSP2;
 
   private final DoubleSolenoid intakePistonL = new DoubleSolenoid(Constants.INTAKE_PISTON_PCM_CHAN_LF_DEPLOY,
       Constants.INTAKE_PISTON_PCM_CHAN_LF_STOW);
```
terbos90803: There is only 1 solenoid for the intake pistons.  Why are 2 defined here?

src/main/java/frc/robot/subsystems/ShooterHoodSubsystem.java
```diff
@@ -8,27 +8,29 @@
...
 
 import frc.robot.Constants;
 import frc.robot.utils.MoPrefs;
-import frc.robot.utils.SimmableCANSparkMax;
+
+import edu.wpi.first.wpilibj.smartdashboard.SmartDashboard;
 
 public class ShooterHoodSubsystem extends SubsystemBase {
-  private final CANSparkMax hoodNEO = new SimmableCANSparkMax(Constants.SPARKMAX_SHOOTER_HOOD_CAN_ADDR,
+  private final CANSparkMax hoodNEO = new CANSparkMax(Constants.SPARKMAX_SHOOTER_HOOD_CAN_ADDR,
```
jordan-powers: What did I miss? The SimmableCANSparkMax should have been a seamless replacement

src/main/java/frc/robot/subsystems/IntakeSubsystem.java
```diff
@@ -40,26 +42,38 @@
   public IntakeSubsystem(MoPDP pdp) {
     intakeSP = new SafeSP(Constants.INTAKE_VICTORSP_PWM_CHAN, SAFE_SPEED, SAFE_COOLDOWN_MS, pdp
         .MakeOvercurrentMonitor(Constants.INTAKE_VICTORSP_PDP_CHAN, UNSAFE_CURRENT_LIMIT, UNSAFE_CURRENT_TIMEOUT_MS));
+    intakeSP2 = new SafeSP(Constants.INTAKE_VICTORSP_PWM_CHAN_2, SAFE_SPEED, SAFE_COOLDOWN_MS, pdp
+        .MakeOvercurrentMonitor(Constants.INTAKE_VICTORSP_PDP_CHAN, UNSAFE_CURRENT_LIMIT, UNSAFE_CURRENT_TIMEOUT_MS));
+
+    intakeSP.setInverted(true);
+    intakeSP2.setInverted(false);
```
terbos90803: Need to validate that intake motors are wired red-red/black-black and adjust these settings if necessary for minimum future confusion.

src/main/java/frc/robot/subsystems/ShooterHoodSubsystem.java
```diff
@@ -60,27 +62,35 @@ public ShooterHoodSubsystem() {
     hoodNEO.setIdleMode(IdleMode.kBrake);
     // The Hood needs to run in the negative direction *towards* the limit switch
     // If it runs the wrong way, flip this invert setting.
-    hoodNEO.setInverted(false);
+    hoodNEO.setInverted(true);
 
     reliableZero = false;
     stopHood();
+
+    CommandScheduler.getInstance().registerSubsystem(this);
```
jordan-powers: This is handled in the constructor for the supertype SubsystemBase, so you shouldn't do it here.

src/main/java/frc/robot/subsystems/ShooterHoodSubsystem.java
```diff
@@ -60,27 +62,35 @@ public ShooterHoodSubsystem() {
...
+  /*
+   * public void raiseShotAngle() { ++hoodSetpoint; }
+   * 
+   * public void lowerShotAngle() { --hoodSetpoint; }
+   */
+
   public void moveHood(double posRequest) {
     // Used for autonomous and vision-tied control of the shooter hood.
-    hoodPID.setReference(posRequest, ControlType.kSmartMotion, 0);
+    hoodPID.setReference(posRequest, ControlType.kPosition, 0);
```
jordan-powers, line 73: Why kPosition instead of kSmartMotion?

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -90,62 +105,119 @@ public ShooterSubsystem(ShooterHoodSubsystem shooterHood) {
...
+  }
+
+  private void updateVelocity() {
+    double currPos = shooterEncoder.getPosition();
+    double currTime = Timer.getFPGATimestamp();
+    double dPos = currPos - lastPos;
+    double dT = currTime - lastTime;
+    lastPos = currPos;
+    lastTime = currTime;
+    currVelocity = dPos / dT * 60;
```
jordan-powers: What's wrong with `shooterEncoder.getVelocity()` ([docs](http://www.revrobotics.com/content/sw/max/sw-docs/java/com/revrobotics/CANEncoder.html#getVelocity()))?

src/main/java/frc/robot/subsystems/StorageSubsystem.java
```diff
@@ -11,8 +11,8 @@
 
 public class StorageSubsystem extends SubsystemBase {
 
-  private final double SAFE_SPEED = 0;
-  private final int SAFE_COOLDOWN_MS = 1000;
+  private final double SAFE_SPEED = 1;
+  private final int SAFE_COOLDOWN_MS = 0;
```
jordan-powers: This pretty much invalidates the entire purpose of the SafeSP class. If you don't want the current monitoring, just use the VictorSP class.

src/main/java/frc/robot/subsystems/conditioners/CurvesConditioner.java
```diff
@@ -3,8 +3,8 @@
 import org.usfirst.frc.team4999.utils.Utils;
 
 public class CurvesConditioner implements DriveConditioner {
-  private final double move_curve = 2;
-  private final double turn_curve = 2;
+  private final double move_curve = 1;
+  private final double turn_curve = 1;
```
jordan-powers: I assume this was at the driver's request.
What problem were you trying to solve?
This shouldn't affect wheelies, in theory it actually makes the robot harder to control.

src/main/java/frc/robot/utils/MoPrefs.java
```diff
@@ -58,7 +57,7 @@ public static double getShooterHoodSetpoint() {
   }
 
   public static double getShooterHoodPositionTolerance() {
-    return getDouble("SHOOTER_HOOD_POSITION_TOLERANCE", SHOOTER_HOOD_POSITION_TOLERANCE);
+    return 2;
```
jordan-powers: Umm, what's the point of using the preferences if you're just going to adjust it in code?

src/main/java/frc/robot/subsystems/FalconDriveSubsystem.java
```diff
@@ -153,8 +154,8 @@ public void drive(final double moveRequest, final double turnRequest) {
       leftFront.set(ControlMode.Velocity, leftETPerS);
       rightFront.set(ControlMode.Velocity, rightETPerS);
     } else {
```
jordan-powers, line 165: You should work on getting PID working on the drive.
That's the way to prevent wheelies since you can set the max acceleration as part of the Smart Motion.

src/main/java/frc/robot/subsystems/Limelight.java
```diff
@@ -0,0 +1,114 @@
...
+  public static final double RANGE_Y = 24.85;
+
+  public static final double TARGET_DIST = 3;
+  public static final double DIST_ERR = 0;
+  public static final double X_ERR = 0;
+  public static final double Y_ERR = 0;
+
+  private static final double CAMERA_ANGLE = 56.5;
+  private static final double CAMERA_HEIGHT = 23.6;
+  private static final double TARGET_HEIGHT = 84;
```
terbos90803: These are specific to our robot.  They should be documented and likely moved to the Constants file.

At some point, the Limelight code should be packaged as a reusable utility with appropriate configuration params.

src/main/java/frc/robot/commands/DriveToWall.java
```diff
@@ -0,0 +1,39 @@
...
+
+public class DriveToWall extends CommandBase {
+  private final FalconDriveSubsystem m_subsystem;
+  private final ShooterSubsystem m_shooter;
+  private final StorageSubsystem m_storage;
+
+  public DriveToWall(FalconDriveSubsystem subsystem, ShooterSubsystem shooter, StorageSubsystem storage) {
+    m_subsystem = subsystem;
+    m_shooter = shooter;
+    m_storage = storage;
```
jordan-powers: You're not requiring the subsystems you're using

src/main/java/frc/robot/controllers/ControllerBase.java
jordan-powers, line 32: It feels confusing that half of the inputs are defined in the RobotContainer and half of the inputs are defined in the DriveController.
This is why in previous years we didn't use JoystickButtons, instead using [BooleanTriggers](https://github.com/momentumfrc/2019DeepSpace/blob/1c7e3cb6882befb5f5d0f91fd71fd124007be35b/2019RobotCodeDeepSpace/src/main/java/frc/robot/OI.java#L54-L60) linked to the DriveController class.

src/main/java/frc/robot/subsystems/Limelight.java
jordan-powers: Have these been calibrated?
How variable are they, should they be in MoPrefs?

src/main/java/frc/robot/subsystems/Limelight.java
```diff
@@ -0,0 +1,114 @@
...
+    private final double yCoord;
+    private final boolean validTarget;
+    private final double dist;
+
+    private LimelightData(double valid, double xAngle, double yAngle) {
+      this.xCoord = xAngle;
+      this.yCoord = yAngle;
+
+      double height = TARGET_HEIGHT - CAMERA_HEIGHT;
+      double slope = Math.tan(CAMERA_ANGLE + yAngle);
```
terbos90803: Units are not correct?  Math.tan is in radians, but those angles are clearly in degrees

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -66,19 +71,29 @@
...
+  private double lastPos = 0;
+  private double lastTime = 0;
+  private double currVelocity = 0;
 
-  public ShooterSubsystem(ShooterHoodSubsystem shooterHood) {
+  public ShooterSubsystem(final ShooterHoodSubsystem shooterHood) {
 
     this.shooterHood = shooterHood;
 
+    NetworkTableInstance.getDefault().getTable("limelight").getEntry("ledMode").setNumber(3);
```
jordan-powers: Hmm, this [smells](https://en.wikipedia.org/wiki/Code_smell)

src/main/java/frc/robot/subsystems/Limelight.java
```diff
@@ -0,0 +1,114 @@
...
+    public boolean valid() {
+      return validTarget;
+    }
+
+    public boolean targetMet() {
+      if (valid()) {
+        double x = Math.abs(xCoord());
+        double y = Math.abs(yCoord());
+        double d = Math.abs(dist() - DIST_ERR);
+        if (x <= X_ERR && y <= Y_ERR && d <= TARGET_DIST) {
```
terbos90803: It hardly seems likely that we can match all 3 at the same time.  In fact, y and d are coupled, so only one should be tested.

src/main/java/frc/robot/subsystems/Limelight.java
```diff
@@ -0,0 +1,114 @@
...
+
+    public double yCoord() {
+      return yCoord;
+    }
+
+    public boolean valid() {
+      return validTarget;
+    }
+
+    public boolean targetMet() {
```
terbos90803, line 112: I question whether targetMet is living in the correct place.  I don't see an "easier" place to move it at the moment, but that might pop up at some point.

Wherever targetMet ends up living, we should consider a companion method to generate move and turn requests based on the same formulas so we can feed that directly into the drive PID.

src/main/java/frc/robot/subsystems/ShooterHoodSubsystem.java
terbos90803: Is there a reason why SimmableCANSparkMax doesn't work here?

src/main/java/frc/robot/subsystems/ShooterHoodSubsystem.java
```diff
@@ -8,27 +8,29 @@
...
 
-  private final double kP = 5e-5;
+  private final double kP = 0.15;
   private final double kI = 1e-6;
   private final double kD = 0;
   private final double kIz = 0;
-  private final double kFF = 0.000156;
-  private final double kMaxOutput = 0.3;
-  private final double kMinOutput = -0.3;
+  private final double kFF = 0;
```
terbos90803: kFF should be tuned first, then kP, then kI if necessary.  I suspect we won't need any kI or kD for this one.

src/main/java/frc/robot/subsystems/ShooterHoodSubsystem.java
```diff
@@ -8,27 +8,29 @@
...
+  private final double kMaxOutput = 1.0;
+  private final double kMinOutput = -1.0;
```
terbos90803: We'll always have balanced outputs here, so can collapse to a single constant that's negated as needed.

jordan-powers (commented):
I can't request changes since I technically opened the PR. But I am requesting changes.

Also, I can't leave a comment on the file `getDefault.java` because it is a completely blank file which has been created for some reason.

src/main/java/frc/robot/subsystems/ShooterHoodSubsystem.java
```diff
@@ -60,27 +62,35 @@ public ShooterHoodSubsystem() {
...
+    CommandScheduler.getInstance().registerSubsystem(this);
   }
 
+  /*
+   * public void raiseShotAngle() { ++hoodSetpoint; }
+   * 
+   * public void lowerShotAngle() { --hoodSetpoint; }
+   */
+
   public void moveHood(double posRequest) {
```
terbos90803: Might change the name to setHoodPosition or something other than move

src/main/java/frc/robot/subsystems/ShooterHoodSubsystem.java
```diff
@@ -93,16 +103,23 @@ public void stopHood() {
...
   }
 
   @Override
   public void periodic() {
     hoodPos = hoodEncoder.getPosition();
-    if (hoodLimitSwitch.get()) {
-      zeroHood();
-      reliableZero = true;
+    SmartDashboard.putNumber("pose", hoodPos);
+    if (!reliableZero) {
```
terbos90803: Why condition checking the limit on reliableZero?  It should probably be checked constantly.  Did you encounter a specific problem around this?

src/main/java/frc/robot/subsystems/ShooterHoodSubsystem.java
```diff
@@ -60,27 +62,35 @@ public ShooterHoodSubsystem() {
...
-      hoodPID.setReference(MoPrefs.getShooterHoodSetpoint(), ControlType.kSmartMotion, 0);
+      hoodPID.setReference(MoPrefs.getShooterHoodSetpoint(), ControlType.kPosition, 0);
     else
       hoodNEO.set(SAFE_STOW_SPEED);
   }
 
   public void stowHood() {
     if (reliableZero)
-      hoodPID.setReference(0, ControlType.kSmartMotion, 0);
+      hoodPID.setReference(0, ControlType.kPosition, 0);
```
terbos90803, line 88: Not sure if we should stow at 0.  I was thinking it should be somewhere just above the limit switch so you never touch the limit switch during normal operation.
Alternately, the stow position should *always* rely on the limit switch and not a specific position.

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -90,62 +105,119 @@ public ShooterSubsystem(ShooterHoodSubsystem shooterHood) {
...
+
+    if (enablePID) {
+      shooterPIDRight.setReference(MoPrefs.getShooterPIDSetpoint(), ControlType.kVelocity);
+    } else {
+      leader_shooterMAXRight.set(MoPrefs.getShooterFlywheelSetpoint());
+    }
+
+    final boolean shooterHoodReady = shooterHood.hasReliableZero() && shooterHood.getFullyDeployed();
+    final boolean shooterWheelReady = MoPrefs.getShooterFlywheelSetpoint()
+        - leader_shooterMAXRight.getEncoder().getVelocity() < MoPrefs.getShooterFlywheelTolerance();
```
terbos90803: Using the wrong getVelocity and the diff needs to be Math.abs()

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -90,62 +105,119 @@ public ShooterSubsystem(ShooterHoodSubsystem shooterHood) {
...
+    if (shooterHoodReady && shooterWheelReady) {
+      shooterGate.set(MoPrefs.getShooterGateSetpoint());
+    } else {
+      shooterGate.set(0);
+    }
+
+  }
+
+  public void shootAuto() {
```
terbos90803: Is there a particular reason why shootAuto() is different than shoot()?  Seems like a maintenance problem with no benefit.

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -90,62 +105,119 @@ public ShooterSubsystem(ShooterHoodSubsystem shooterHood) {
...
+
+    if (maintainFlywheelAtIdle) {
+      if (enablePID) {
+        shooterPIDRight.setReference(MoPrefs.getShooterFlywheelIdle(), ControlType.kVelocity);
+      } else {
+        leader_shooterMAXRight.set(MoPrefs.getShooterFlywheelIdle());
+      }
+    } else {
+      if (enablePID) {
+        shooterPIDRight.setReference(0, ControlType.kVelocity);
```
terbos90803: Probably want to use the same stopMotor for both PID and non-PID cases to take advantage of coasting instead of braking.  To be clear, the if should be eliminated and the cases collapsed.

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -90,62 +105,119 @@ public ShooterSubsystem(ShooterHoodSubsystem shooterHood) {
...
   }
 
   public void purge() {
     // stow hood
     // reverse gate
     // reverse shooter wheel
-    shooterHood.stowHood();
-    shooterPIDRight.setReference(-MoPrefs.getShooterFlywheelIdle(), ControlType.kVelocity);
-    shooterGate.set(-MoPrefs.getShooterGateSetpoint());
+    if (enablePID) {
```
terbos90803: No need for PID here.  Always use dumb speed.

No need for -1 * either, just use -
-MoPrefs.getShooterFlywheelIdle()

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -90,62 +105,119 @@ public ShooterSubsystem(ShooterHoodSubsystem shooterHood) {
...
-   * the NEO's velocity. Intended to be called when a button is pressed.
-   */
   public void shoot() {
-    // extend hood
     // fast shooter wheel
     // run gate if both of "" are good
+    // NetworkTableInstance.getDefault().getTable("limelight").getEntry("ledMode").setNumber(3);
+
+    if (enablePID) {
+      shooterPIDRight.setReference(MoPrefs.getShooterPIDSetpoint(), ControlType.kVelocity);
```
terbos90803: Opportunity to coalesce PID and non-PID setpoints.  The ranges are mappable so there should only be one constant to control both.  Otherwise a maintenance problem.

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
terbos90803: If this continues to be noisy, might need to add a median filter to reject bad samples.

src/main/java/frc/robot/subsystems/StorageSubsystem.java
```diff
@@ -22,23 +22,25 @@
...
   }
 
   public void run() {
     speed = MoPrefs.getStorageSpeed();
+    storage.set(speed);
   }
 
   public void reverse() {
-    speed = -MoPrefs.getStorageSpeed();
+    speed = -1 * MoPrefs.getStorageSpeed();
```
terbos90803: Why -1* instead of just -?

src/main/java/frc/robot/subsystems/StorageSubsystem.java
```diff
@@ -22,23 +22,25 @@
   public StorageSubsystem(MoPDP pdp) {
     storage = new SafeSP(Constants.STORAGE_VICTORSP_PWM_CHAN, SAFE_SPEED, SAFE_COOLDOWN_MS, pdp
         .MakeOvercurrentMonitor(Constants.STORAGE_VICTORSP_PDP_CHAN, UNSAFE_CURRENT_LIMIT, UNSAFE_CURRENT_TIMEOUT_MS));
-    stop();
+    storage.setInverted(false);
   }
 
   public void run() {
     speed = MoPrefs.getStorageSpeed();
+    storage.set(speed);
```
terbos90803: Why set class member speed?  Just call set() with the value.

src/main/java/frc/robot/subsystems/conditioners/ReverseConditioner.java
```diff
@@ -1,7 +1,7 @@
 package frc.robot.subsystems.conditioners;
 
 public class ReverseConditioner implements DriveConditioner {
-  private boolean reversed = false;
+  private boolean reversed = true;
```
terbos90803: Why does reversed default to true?  It should default to false and the backwards robot should be fixed elsewhere.

src/main/java/frc/robot/utils/MoPrefs.java
```diff
@@ -3,19 +3,18 @@
 import edu.wpi.first.wpilibj.Preferences;
 
 public class MoPrefs {
-  static final double INTAKE_ROLLER_SETPOINT = 1.0;
-  /**
-   * The maximum accleration of the intake roller, in motor setpoint per scheduler
-   * cycle.
-   */
+  static final double INTAKE_ROLLER_SETPOINT = -0.125;
```
terbos90803: Wow, we're really slowing down the intake.  I'm thinking we chose the wrong gearboxes.

Also, this should be a positive value.  If the motors run backwards, they should be inverted in the subsystem.

src/main/java/frc/robot/utils/MoPrefs.java
```diff
@@ -3,19 +3,18 @@
...
-  static final double SHOOTER_HOOD_POSITION_TOLERANCE = 0.01;
+  static final double SHOOTER_HOOD_SETPOINT = 100;
+  static final double SHOOTER_HOOD_POSITION_TOLERANCE = 2;
   static final double SHOOTER_GATE_SETPOINT = 1;
   static final double SHOOTER_FLYWHEEL_SETPOINT = 1;
+  static final double SHOOTER_PID_SETPOINT = 4500;
   static final double SHOOTER_FLYWHEEL_IDLE = 0.2;
-  static final double STORAGE_SPEED = 0.5;
+  static final double STORAGE_SPEED = 0.75;
+  static final double SHOOTER_FLYWHEEL_TOLERANCE = 100; // RPM
```
terbos90803, line 14: Might want to tighten up this tolerance.  100 RPM could make a big difference in placement.  I seem to recall that the PID control was pretty tight, so we should be able to get this down to 10ish.  Testing required.

src/main/java/frc/robot/utils/MoPrefs.java
```diff
@@ -3,19 +3,18 @@
...
+  static final double AUTO_CHOOSER = 0;
```
terbos90803: What's this for?  AutoChooser should be a named pulldown, not numbers in the Prefs.

src/main/java/frc/robot/utils/MoPrefs.java
terbos90803: Why is this hardcoded?  Should be getting from prefs.

src/main/java/frc/robot/utils/MoPrefs.java
```diff
@@ -74,10 +73,22 @@ public static double getShooterFlywheelIdle() {
   }
 
   public static double getIntakeRollerAccRamp() {
-    return getDouble("INTAKE_ROLLER_ACC_RAMP", INTAKE_ROLLER_ACC_RAMP);
+    return 0.1;
```
terbos90803: same

terbos90803 (changes requested):

src/main/java/frc/robot/commands/ShootFromLine.java
terbos90803: Not on this line.  He's trying to express 13 > time > 5

src/main/java/frc/robot/subsystems/ShooterHoodSubsystem.java
terbos90803: kPosition is easier to get working correctly and is sufficient for a low-inertia system like the shooter hood.  It works just like a COTS servo in this mode.

src/main/java/frc/robot/subsystems/StorageSubsystem.java
jordan-powers: Personally, I find it more legible. The `-` can be easy to miss.

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
terbos90803: getVelocity was returning the same value as getPosition.  I have a CD thread open on this topic.  https://www.chiefdelphi.com/t/sparkmax-java-encoder-getvelocity-returning-position-value/382020

src/main/java/frc/robot/subsystems/conditioners/CurvesConditioner.java
terbos90803: This is what I said.  I definitely saw a change in the behavior after they made this change, but it could have been as much from driver practice as from a math change.  Needs more testing.

src/main/java/frc/robot/subsystems/Limelight.java
terbos90803: I think RANGE_X and RANGE_Y are constants of the Limelight.  All the others are based on our robot.

src/main/java/frc/robot/utils/SimmableCANSparkMax.java
```diff
@@ -50,6 +59,78 @@ public CANError setReference(double value, ControlType ctrl, int pidSlot, double
...
+
+    @Override
+    public double getVelocity() {
+      if (m_simDevice != null) {
+        if (m_encoderCount == null) {
+          m_encoderCount = m_simDevice.createDouble("Encoder Vel", false, 0);
+        }
+        return m_encoderCount.get();
+      } else {
+        return super.getPosition();
```
sciencewhiz: Should return super.getVelocity().

terbos90803: And there it is.  Thanks so much for tracking that one down.

jordan-powers: Ah, whoops. That one's on me

src/main/java/frc/robot/utils/SimmableCANSparkMax.java
```diff
@@ -125,7 +125,7 @@ public double getVelocity() {
         }
         return m_encoderCount.get();
```
terbos90803: This seems wrong too.  Position and Velocity should be different things.

terbos90803 (changes requested):

src/main/java/frc/robot/utils/SimmableCANSparkMax.java
jordan-powers: Looks like I just copy-pasted the code from encoderCount and updated the
string without updating any of the actual code. Oops.

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -98,33 +89,32 @@
...
   private final DriveCommand driveCommand = new DriveCommand(falconDriveSubsystem, mainController, driveConditioner);
 
   private final Command shootCommand = new RunCommand(shooterHoodSubsystem::deployHood, shooterHoodSubsystem)
       .alongWith(new RunCommand(shooterSubsystem::shoot, shooterSubsystem))
       .alongWith(new RunCommand(limelight::lightsOn, limelight));
 
-  private final Command shootIdleAutoCommand = new RunCommand(shooterSubsystem::idle, shooterSubsystem)
-      .alongWith(new RunCommand(shooterHoodSubsystem::stowHood, shooterHoodSubsystem));
+  private final Command shootFromLine = new ParallelCommandGroup(
+      new RunCommand(shooterSubsystem::shoot, shooterSubsystem).withTimeout(10),
```
terbos90803: I'm not sure I'm fully understanding the intent of this command group.  An English description would be outstanding.

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -98,33 +89,32 @@
...
+      new RunCommand(shooterHoodSubsystem::deployHood, shooterHoodSubsystem).withTimeout(10),
+      new RunCommand(intakeSubsystem::runIntakeFwd, intakeSubsystem).withTimeout(10)
+          .andThen(new RunCommand(() -> falconDriveSubsystem.drive(0.5, 0), falconDriveSubsystem)));
 
-  private final ShootFromLine shootFromLine = new ShootFromLine(falconDriveSubsystem, shooterSubsystem,
-      storageSubsystem, intakeSubsystem, shooterHoodSubsystem);
+  private final Command shootFromWall = new ParallelCommandGroup(
```
terbos90803: Same here.  I'm not understanding the use of the ParallelCommandGroup to achieve a sequence of steps.

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -98,33 +89,32 @@
...
+      new RunCommand(() -> falconDriveSubsystem.drive(0.5, 0)).withTimeout(5).andThen(
+          new RunCommand(shooterSubsystem::shoot, shooterSubsystem),
+          new RunCommand(() -> shooterHoodSubsystem.setHoodPosition(60), shooterHoodSubsystem),
```
terbos90803: setHoodPosition(magic_number).

I still don't understand why the shooterHood is treated differently from the shooter.  There should be a hierarchy of subsystems such that if you want to shoot, you send a command to the shooter and it figures out what to do with the hood to make the shots correctly.  The top level command shouldn't need to know about both.

src/main/java/frc/robot/subsystems/Limelight.java
```diff
@@ -71,8 +75,8 @@ private LimelightData(double valid, double xAngle, double yAngle) {
       this.xCoord = xAngle;
       this.yCoord = yAngle;
 
-      double height = TARGET_HEIGHT - CAMERA_HEIGHT;
-      double slope = Math.tan(CAMERA_ANGLE + yAngle);
+      double height = Constants.TARGET_HEIGHT - Constants.CAMERA_HEIGHT;
+      double slope = Math.tan(Math.toRadians(Constants.CAMERA_ANGLE) + Math.toRadians(yAngle));
```
terbos90803: Instead of calling toRadians twice, add the angles as degrees and then convert to radians once.

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -45,50 +49,45 @@
...
   /**
    * The Differential Gain of the SparkMAX PIDF controller. The weight of the
    * differential path against the proportional and integral paths is controlled
    * by this value.
    */
-  private final double kD = 0;
+  private static final double kD = 0;
   /**
    * The Integral Zone of the SparkMAX PIDF controller. The integral accumulator
    * will reset once it hits this value.
```
terbos90803, line 67: The integral accumulator will reset if the error exceeds this value.  0 is disabled.

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -121,23 +120,22 @@ public ShooterSubsystem(final ShooterHoodSubsystem shooterHood) {
...
-    lastPos = shooterEncoder.getPosition();
-    lastTime = Timer.getFPGATimestamp();
+    this.tab = tab;
   }
 
   public void shoot() {
     // fast shooter wheel
     // run gate if both of "" are good
-
+    double dumbSetpoint = Utils.map(MoPrefs.getShooterPIDSetpoint(), -5500, 5500, -1, 1);
```
terbos90803: magic numbers.

Instead of "dumb", perhaps "openLoop" is a better name choice.  There may be other options, but I would prefer an accurate description versus op-ed.

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -146,73 +144,29 @@ public void shoot() {
...
-  }
-
   @Override
   public void periodic() {
-    updateVelocity();
-    SmartDashboard.putNumber("Flywheel Speed", getCurrVelocity());
+    SmartDashboard.putNumber("Flywheel Speed", shooterEncoder.getVelocity());
     SmartDashboard.putNumber("Flywheel Position", shooterEncoder.getPosition());
+
+    NetworkTableEntry shooterPIDchooser = tab.add("Shooter PID", true).withWidget(BuiltInWidgets.kToggleSwitch)
```
terbos90803: This can't be in periodic().  Adding the chooser can only happen once so it should be in init

terbos90803 (changes requested):

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
FireShredder24: The default NEO constants taken from last year's code have the IZ as 0, so that's what I went off of.

src/main/java/frc/robot/subsystems/ShooterHoodSubsystem.java
```diff
@@ -69,12 +71,14 @@ public ShooterHoodSubsystem() {
...
     hoodPID.setReference(posRequest, ControlType.kPosition, 0);
+    currSetpoint = posRequest;
   }
 
   public void deployHood() {
-    if (reliableZero)
+    if (reliableZero) {
       hoodPID.setReference(MoPrefs.getShooterHoodSetpoint(), ControlType.kPosition, 0);
-    else
+      currSetpoint = MoPrefs.getShooterHoodSetpoint();
```
terbos90803, line 79: I'd probably swap the order of these two lines and avoid calling MoPrefs twice.  Just call once and reuse currSetpoint in the setReference call.

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -141,7 +151,29 @@ public void shoot() {
     } else {
       shooterGate.set(0);
     }
+  }
+
+  public void shootFromWall() {
+    // fast flywheel
+    // deploy hood at lower angle
+    // run gate if both of "" are good
+    double openLoopSetpoint = Utils.map(MoPrefs.getShooterPIDSetpoint(), -5500, 5500, -1, 1);
```
terbos90803: This pattern has appeared multiple times.  Please refactor into a single function that eliminates duplicated code.  Also do something about the magic numbers.

terbos90803 (changes requested):

src/main/java/frc/robot/commands/AutonDriveCommand.java
```diff
@@ -56,7 +56,7 @@ public void execute() {
       met = false;
     }
     System.out.format("Target Distance:%.02f\n", distance);
-    drive_subsystem.drive(moveRequest, turnRequest);
+    drive_subsystem.drive(moveRequest * 0, turnRequest);
```
terbos90803: Needs a comment why this is times 0.  Is it a placeholder?  Could it just be 0?  Why?

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -178,19 +160,22 @@ public void shootFromWall() {
...
     shooterGate.stopMotor();
     leader_shooterMAXRight.stopMotor();
   }
 
   public void purge() {
-    // stow hood
     // reverse gate
     // reverse shooter wheel
     leader_shooterMAXRight.set(-0.2);
     shooterGate.set(-1 * MoPrefs.getShooterGateSetpoint());
```
terbos90803, line 172: Why is this -1 here?  Could be just a - in front of MoPrefs...

terbos90803 (changes requested):
Getting really close.

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
FireShredder24: I believe Jordan mentioned earlier that a "-1 *" is far more readable than just a "-".  I can definitely see myself completely missing that character in a refactor scenario.

terbos90803 (approved):
I think we're there.  I'm ready to merge.

PR #24, opened 2021-01-07
Migrate repo to WPILib 2021.1.2

FireShredder24:


terbos90803 (commented):
This all looks reasonable.

Does anyone understand why some of the diffs don't appear to change the contents of the file, yet the whole file is flagged?  I don't see any visible changes to spacing or anything.

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -176,7 +176,8 @@ private void configureButtonBindings() {
    * @return the command to run in autonomous
    */
   public Command getAutonomousCommand() {
-    return autoChooser.getSelected();
```
jordan-powers: Is this part of the migration, or is this an out-of-scope change which slipped in?

jordan-powers:
I'm guessing the line endings changed. That would cause the diffs appear to show the whole file changing, with no actual substantial changes.

src/main/java/frc/robot/RobotContainer.java
FireShredder24: I included this out of scope to test the autonomous command.  I didn't intend for it to be pushed at all, but apparently I didn't clean the working tree before committing.

.vscode/settings.json
```diff
@@ -1,19 +1,17 @@
...
-    "**/CVS": true,
-    "**/.DS_Store": true,
-    "bin/": true,
-    "**/.classpath": true,
-    "**/.project": true,
-    "**/.settings": true,
-    "**/.factorypath": true
-  },
-  "editor.tabSize": 2,
-  "editor.insertSpaces": true,
```
terbos90803, line 16: I think we should retain these editor settings.  Makes for happier merges.

terbos90803 (changes requested):

terbos90803 (approved):
Looks much better.

PR #26, opened 2021-01-31
Shuffleboard PID tuning widgets

FireShredder24:
Add Shuffleboard PID widgets for drivetrain and shooter
Add more status widgets for shooter hood, shooter, and drivetrain

### DIFF
--- a/src/main/java/frc/robot/RobotContainer.java
+++ b/src/main/java/frc/robot/RobotContainer.java
@@ -60,25 +60,22 @@ public class RobotContainer {
   private final ControllerBase mainController = new ControllerBase(xbox, f310);
 
   // ----------------------------------------Buttons------------------------------------------------
-  private final JoystickButton intakeRollerFwdButton = new JoystickButton(f310, LogitechF310.Button.kBumperLeft.value); // Left
-                                                                                                                        // Bumper
-  private final JoystickButton intakeRollerRvsButton = new JoystickButton(f310, LogitechF310.Button.kBumperRight.value); // Right
-                                                                                                                         // Bumper
-  private final JoystickButton intakePistonToggle = new JoystickButton(f310, LogitechF310.Button.kB.value); // B
+  private final JoystickButton intakeRollerFwdButton = new JoystickButton(f310, LogitechF310.Button.kBumperLeft.value);
+  private final JoystickButton intakeRollerRvsButton = new JoystickButton(f310, LogitechF310.Button.kBumperRight.value);
+  private final JoystickButton intakePistonToggle = new JoystickButton(f310, LogitechF310.Button.kB.value);
 
   private final JoystickButton climberStow = new JoystickButton(f310, 7); // TODO Pick a button and update number
   private final JoystickButton climberClimb = new JoystickButton(f310, 8); // TODO Pick a button and update number
 
-  private final JoystickButton shooterShoot = new JoystickButton(xbox, XboxController.Button.kBumperRight.value); // Right
-                                                                                                                  // Bumper
-  private final JoystickButton purge = new JoystickButton(xbox, XboxController.Button.kBumperLeft.value); // Left bumper
+  private final JoystickButton shooterShoot = new JoystickButton(xbox, XboxController.Button.kBumperRight.value);
+  private final JoystickButton purge = new JoystickButton(xbox, XboxController.Button.kBumperLeft.value);
 
-  private final JoystickButton spdLimitInc = new JoystickButton(xbox, XboxController.Button.kY.value); // Y
-  private final JoystickButton spdLimitDec = new JoystickButton(xbox, XboxController.Button.kA.value); // A
+  private final JoystickButton spdLimitInc = new JoystickButton(xbox, XboxController.Button.kY.value);
+  private final JoystickButton spdLimitDec = new JoystickButton(xbox, XboxController.Button.kA.value);
 
-  private final JoystickButton reverseRobot = new JoystickButton(xbox, XboxController.Button.kB.value); // B
+  private final JoystickButton reverseRobot = new JoystickButton(xbox, XboxController.Button.kB.value);
 
-  private final JoystickButton storageButton = new JoystickButton(f310, LogitechF310.Button.kX.value); // X
+  private final JoystickButton storageButton = new JoystickButton(f310, LogitechF310.Button.kX.value);
   // ----------------------------------------Conditioners--------------------------------------
   private final SpeedLimitConditioner speedLimitConditioner = new SpeedLimitConditioner();
   private final ReverseConditioner reverseConditioner = new ReverseConditioner();
--- a/src/main/java/frc/robot/RobotContainer.java
+++ b/src/main/java/frc/robot/RobotContainer.java
@@ -83,10 +83,10 @@ public class RobotContainer {
       new CurvesConditioner(), reverseConditioner, speedLimitConditioner);
 
   // ---------------------------------------Subsystems----------------------------------------
-  private final FalconDriveSubsystem falconDriveSubsystem = new FalconDriveSubsystem(testTab);
+  private final FalconDriveSubsystem falconDriveSubsystem = new FalconDriveSubsystem(matchTab);
   private final IntakeSubsystem intakeSubsystem = new IntakeSubsystem(powerDistributionPanel);
   private final ClimberSubsystem climberSubsystem = new ClimberSubsystem();
-  private final ShooterHoodSubsystem shooterHoodSubsystem = new ShooterHoodSubsystem();
+  private final ShooterHoodSubsystem shooterHoodSubsystem = new ShooterHoodSubsystem(matchTab);
   private final Limelight limelight = new Limelight(limelightTab, matchTab, 0, 0);
   private final ShooterSubsystem shooterSubsystem = new ShooterSubsystem(shooterHoodSubsystem, matchTab);
   private final StorageSubsystem storageSubsystem = new StorageSubsystem(powerDistributionPanel);
--- a/src/main/java/frc/robot/RobotContainer.java
+++ b/src/main/java/frc/robot/RobotContainer.java
@@ -81,7 +81,7 @@ public class RobotContainer {
   private final ReverseConditioner reverseConditioner = new ReverseConditioner();
   private final DriveConditioner driveConditioner = new ComposedConditioner(new DeadzoneConditioner(),
       new CurvesConditioner(), reverseConditioner, speedLimitConditioner,
-      new FunctionalConditioner((mr) -> mr, (tr) -> -tr));
+      new FunctionalConditioner(mr -> mr, tr -> -tr));
 
   // ---------------------------------------Subsystems----------------------------------------
   private final FalconDriveSubsystem falconDriveSubsystem = new FalconDriveSubsystem(matchTab);
--- a/src/main/java/frc/robot/RobotContainer.java
+++ b/src/main/java/frc/robot/RobotContainer.java
@@ -80,7 +80,8 @@ public class RobotContainer {
   private final SpeedLimitConditioner speedLimitConditioner = new SpeedLimitConditioner();
   private final ReverseConditioner reverseConditioner = new ReverseConditioner();
   private final DriveConditioner driveConditioner = new ComposedConditioner(new DeadzoneConditioner(),
-      new CurvesConditioner(), reverseConditioner, speedLimitConditioner);
+      new CurvesConditioner(), reverseConditioner, speedLimitConditioner,
+      new FunctionalConditioner((mr) -> mr, (tr) -> -tr));
 
   // ---------------------------------------Subsystems----------------------------------------
   private final FalconDriveSubsystem falconDriveSubsystem = new FalconDriveSubsystem(matchTab);
--- a/src/main/java/frc/robot/RobotContainer.java
+++ b/src/main/java/frc/robot/RobotContainer.java
@@ -26,6 +26,7 @@ import frc.robot.subsystems.conditioners.CurvesConditioner;
 import frc.robot.subsystems.conditioners.DeadzoneConditioner;
 import frc.robot.subsystems.conditioners.SpeedLimitConditioner;
 import frc.robot.utils.MoPrefs;
+import frc.robot.utils.MoPrefs.MoPrefsKey;
 import frc.robot.controllers.ControllerBase;
 
 import edu.wpi.first.wpilibj.GenericHID;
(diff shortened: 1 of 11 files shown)


src/main/java/frc/robot/RobotContainer.java
```diff
@@ -60,33 +60,36 @@
   private final ControllerBase mainController = new ControllerBase(xbox, f310);
 
   // ----------------------------------------Buttons------------------------------------------------
-  private final JoystickButton intakeRollerFwdButton = new JoystickButton(f310, LogitechF310.Button.kBumperLeft.value);
-  private final JoystickButton intakeRollerRvsButton = new JoystickButton(f310, LogitechF310.Button.kBumperRight.value);
-  private final JoystickButton intakePistonToggle = new JoystickButton(f310, LogitechF310.Button.kB.value);
+  private final JoystickButton intakeRollerFwdButton = new JoystickButton(f310, LogitechF310.Button.kBumperLeft.value); // Left
```
terbos90803: Does the comment add anything here?  Is kBumperLeft not readable?  Adding a comment also adds maintenance.

Same feedback for all the comments added here.

src/main/java/frc/robot/commands/DriveCommand.java
```diff
@@ -35,7 +35,7 @@ public void initialize() {
   @Override
   public void execute() {
     m_subsystem.drive(m_conditioner.conditionMove(m_controller.getMoveRequest()),
-        m_conditioner.conditionTurn(m_controller.getTurnRequest()));
+        m_conditioner.conditionTurn(-m_controller.getTurnRequest()));
```
terbos90803: Why was this needed now?  It was working correctly before.

src/main/java/frc/robot/subsystems/FalconDriveSubsystem.java
```diff
@@ -58,31 +59,44 @@
...
    */
-  private static final double K_FF = 1;
+  private double kFF = 1;
 
   private final DifferentialDriveKinematics kinematics = new DifferentialDriveKinematics(
       Units.inchesToMeters(DRIVE_BASE_WIDTH_INCHES));
 
+  private NetworkTableEntry leftDriveVelocity;
+  private NetworkTableEntry rightDriveVelocity;
+  private NetworkTableEntry kPSlider;
```
terbos90803: There used to be a combined element to show all the PID coeffs together.  Is that still around?

src/main/java/frc/robot/subsystems/ShooterHoodSubsystem.java
```diff
@@ -103,7 +109,8 @@ private boolean hasReliableZero() {
   }
 
   private boolean isFullyDeployed() {
-    return Math.abs(getHoodPos() - currSetpoint) < MoPrefs.getShooterHoodPositionTolerance();
+    return MoPrefs.getShooterHoodSetpoint() - MoPrefs.getShooterHoodPositionTolerance() < getHoodPos()
```
terbos90803: Making tolerance comparisons should always be in the form:
Math.abs(setpoint - currPos) < tolerance

Why did this get changed?

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -165,19 +198,38 @@ public void purge() {
...
   }
 
   private boolean isFlywheelReady() {
-    return Math.abs(MoPrefs.getShooterPIDSetpoint() - shooterEncoder.getVelocity()) < MoPrefs
-        .getShooterFlywheelTolerance();
+    return Math.abs(setpoint - shooterEncoder.getVelocity()) < MoPrefs.getShooterFlywheelTolerance();
+  }
+
+  private void updatePIDConstants() {
+    shooterPIDLeft.setP(kP, 0);
```
terbos90803: You can't have two motors mechanically connected and also PID them separately.  The PID has to be on one of the motors and the other motor follows the first one.

terbos90803 (changes requested):

src/main/java/frc/robot/commands/DriveCommand.java
FireShredder24: Well, turning was inverted.  I don't know why, but it was.  I'd guess that some tweaking to Conditioners over the summer did that.

src/main/java/frc/robot/subsystems/FalconDriveSubsystem.java
FireShredder24: There is, but it doesn't work with Talon FX(or Spark MAX), since neither extends the WPILib PIDController class.

src/main/java/frc/robot/subsystems/ShooterHoodSubsystem.java
FireShredder24: I was having a lot of trouble with the it, so I messed around with the logic.  Eventually, it turned out that getHoodPos() wasn't actually polling the encoder.  I just didn't change it back because we moved on to tuning the shooter flywheel PID.  I'll fix it once I finish my homework.

src/main/java/frc/robot/RobotContainer.java
FireShredder24: In hindsight, it didn't help much.  Will revert after homework is done.

src/main/java/frc/robot/commands/DriveCommand.java
terbos90803: It's much more important to understand why than to just patch until working.  Doing that often leads to a big mess.
For instance, it is normal to invert the raw Y axis because of the long history of video games on legacy computers where Y is positive down on the screen.
It is not normal to invert X, so I'd look into this a lot more closely.

src/main/java/frc/robot/subsystems/FalconDriveSubsystem.java
terbos90803: I thought Jordan wrote a wrapper that didn't rely on that class.  Or am I mis-remembering.

src/main/java/frc/robot/subsystems/ShooterHoodSubsystem.java
terbos90803: Another case of needing better understanding before hacking.
I get that if you aren't familiar with the standard tolerance pattern that it may have seemed odd, but now it seems odd and difficult to read for the rest of us.

As a general rule, any time you have to work to understand a bit of code, that's a screaming flag that it deserves a comment block to explain what's going on and why.  This is doubly true if you need to change something to work around a bug.  When the real bug gets fixed, it needs to be easy to find any workarounds that will no longer behave as expected and fix them too.

src/main/java/frc/robot/subsystems/FalconDriveSubsystem.java
FireShredder24: According to the Java API, the widget kPIDController ([here](https://first.wpi.edu/wpilib/allwpilib/docs/release/java/edu/wpi/first/wpilibj/shuffleboard/BuiltInWidgets.html#kPIDController)) relies on a deprecated version of PIDController.  It's weird.  For now, I've just been using separate widgets, organizing them into lists.  It's not very screen-space-efficient, but it works.

src/main/java/frc/robot/subsystems/ShooterHoodSubsystem.java
```diff
@@ -91,19 +98,24 @@ public void stowHood() {
   }
 
   public double getHoodPos() {
-    return hoodPos;
+    return hoodEncoder.getPosition();
```
terbos90803: Just out of curiosity, why did this move here?  It used to be in periodic().  Unless we truly need to get the hood position more frequently than regular period, it's better to read it once and cache the result.

src/main/java/frc/robot/subsystems/ShooterHoodSubsystem.java
```diff
@@ -112,21 +124,23 @@ public boolean isHoodReady() {
 
   @Override
   public void periodic() {
-    hoodPos = hoodEncoder.getPosition();
```
terbos90803, line 115: This line needs to come back.

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -125,16 +126,46 @@ public ShooterSubsystem(final ShooterHoodSubsystem shooterHood, ShuffleboardTab
...
+
+    // Adds a Shuffleboard widget to show whether the flywheel is spinning within a
+    // certain tolerance of the setpoint. See isFlywheelReady().
+    isFlywheelReady = tab.add("Is Flywheel Ready?", false).withWidget(BuiltInWidgets.kBooleanBox).getEntry();
+
+    shooterSetpoint = tab.addPersistent("Shooter Setpoint", 4500).withWidget(BuiltInWidgets.kTextView).getEntry();
+    shooterSetpoint.addListener(notice -> setpoint = notice.value.getDouble(),
+        EntryListenerFlags.kNew | EntryListenerFlags.kUpdate);
+
+    kPSlider = tab.addPersistent("Shooter kP", 5e-5).withWidget(BuiltInWidgets.kTextView).getEntry();
```
terbos90803: There should be no hardcoded constants in these calls.  Especially since these are tuning values and already exist elsewhere.  Keeping them in sync is high maintenance.

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -125,16 +126,46 @@ public ShooterSubsystem(final ShooterHoodSubsystem shooterHood, ShuffleboardTab
...
+    kFFSlider.addListener(notice -> kFF = notice.value.getDouble(),
+        EntryListenerFlags.kNew | EntryListenerFlags.kUpdate);
   }
 
   public void shoot(double hoodSetpoint) {
     // fast shooter wheel
     // deploy hood
     // run gate if both of "" are good
 
-    double pidSetpoint = MoPrefs.getShooterPIDSetpoint();
```
terbos90803, line 137: What the heck is going on here?  The setpoint should be a MoPref and not hardcoded (aside from a default).

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -165,19 +196,33 @@ public void purge() {
...
   @Override
   public void periodic() {
-    SmartDashboard.putNumber("Flywheel Speed", shooterEncoder.getVelocity());
-    SmartDashboard.putNumber("Flywheel Position", shooterEncoder.getPosition());
-    SmartDashboard.putBoolean("Flywheel ready?", isFlywheelReady());
+    flywheelSpeed.setDouble(shooterEncoder.getVelocity());
+    isFlywheelReady.setBoolean(isFlywheelReady());
+    updatePIDConstants();
+
+    setpoint = shooterSetpoint.getDouble(4500);
```
terbos90803: What's going on here?  Why is the setpoint being updated every periodic and there's a hardcoded constant which is repeated multiple times in the code.

terbos90803 (changes requested):

src/main/java/frc/robot/subsystems/ShooterHoodSubsystem.java
FireShredder24: Fixed

src/main/java/frc/robot/subsystems/ShooterHoodSubsystem.java
FireShredder24: It has

src/main/java/frc/robot/commands/DriveCommand.java
FireShredder24: Jordan fixed this by adding a FunctionalConditioner to the ComposedConditioner that is passed to DriveCommand in RobotContainer.  It reverses turnRequest while leaving moveRequest untouched.

src/main/java/frc/robot/subsystems/FalconDriveSubsystem.java
```diff
@@ -185,4 +159,40 @@ public void stop() {
...
+  }
+
+  // Updates the PID constants stored in each Talon FX from MoPrefs
+  private void updatePIDConstants() {
+    double kP = MoPrefs.getDriveKP();
+    double kI = MoPrefs.getDriveKI();
+    double kD = MoPrefs.getDriveKD();
+    int kIZ = (int) MoPrefs.getDriveKIZ();
+    double kFF = MoPrefs.getDriveKFF();
+    leftFront.config_kP(0, kP);
```
terbos90803, line 181: I know it adds complexity, but I seem to recall that you shouldn't be writing PID constants every period.  You'll have to figure out how to only write them on changes.  That usually requires caching the old value and only writing on diffs.  Jordan can probably help make this look all fancy so the code doesn't become cumbersome.

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -165,19 +139,32 @@ public void purge() {
...
+    return Utils.map(closedLoopSetpoint, -MAX_FREE_SPEED, MAX_FREE_SPEED, -1, 1);
   }
 
   private boolean isFlywheelReady() {
     return Math.abs(MoPrefs.getShooterPIDSetpoint() - shooterEncoder.getVelocity()) < MoPrefs
         .getShooterFlywheelTolerance();
   }
 
+  private void updatePIDConstants() {
+    shooterPIDRight.setP(MoPrefs.getShooterKP(), 0);
```
terbos90803: Only update on change

src/main/java/frc/robot/utils/MoPrefs.java
```diff
@@ -86,4 +109,72 @@ public static double getShooterPIDSetpoint() {
   public static double getShootFromWallHoodSetpoint() {
     return getDouble("Shoot From Wall Hood Setpoint", SHOOT_FROM_WALL_HOOD_SETPOINT);
   }
+
+  public static double getShooterKP() {
```
terbos90803: Does anyone else find this code pattern a challenge to maintain?
Why do we need specific methods baked into a shared class to get values from a dictionary?
It has always seemed to me that this should be easier.  Specifically, MoPrefs should maintain the dictionary, but not know/care about what's in it.  The clients should be responsible for knowing their own keys.

terbos90803 (changes requested):

src/main/java/frc/robot/utils/MoPrefs.java
jordan-powers: Well, MoPrefs is just a wrapper around ` edu.wpi.first.wpilibj.Preferences`. The only functionality it adds is the individual method calls. As far as I recall, these are the reasons for why we chose to create the MoPrefs instead of directly using the Preferences object:
1. To provide explicit default values for each preference
2. To make it very obvious what values are stored in the preferences
3. To prevent errors where typos in a string key ends up getting unexpected values

However, I do agree that this way of achieving that goal is pretty clunky.

Maybe a better way to do this would be to create a `enum MoPrefKeys` and refactor `MoPrefs` to only have one public method which takes a MoPrefKey and returns the value. Java's enums are basically just classes, so we could define a MoPrefKey to hold both the string key and an explicit default value.

That shouldn't be too hard to implement: I'll create a PR.

jordan-powers: See #27

src/main/java/frc/robot/subsystems/FalconDriveSubsystem.java
jordan-powers: This would be easy if the Preferences class allowed us access to the underlying `NetworkTableEntry`s.  Unfortunately, that is not the case.
Also, references is a final class, so we can't subclass it and give ourselves access.

I can see two options:
1. Re-implement the Preferences class, and add a method to get access to the `NetworkTableEntry`. The Preferences class is pretty simple, so this wouldn't be too difficult. It does seem a bit hacky though.
2. Create some sort of "watcher" class that watches a value, looks for changes, and triggers a lambda when the value changes.  This would be pretty trivial to implement, however it would require polling the Preferences NetworkTable on every loop. This is probably not a big deal, though, and I suspect the NetworkTables are designed to support it.

I'm not sure which option is better.

terbos90803: I thought MoPrefs already subclassed Preferences.  Apparently, it does not, so it must wrap it instead (composition vs inheritance).

I also thought there was a way to subscribe to NT entries, but if there's not, then having a mechanism in place to detect and notify on changes would certainly be helpful.

src/main/java/frc/robot/utils/MoPrefs.java
```diff
@@ -47,43 +123,11 @@ public static void setDouble(String key, double value) {
     System.out.format("set pref: %s=%f\n", key, value);
```
aimiebell, line 123: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [FS: Format string should use %n rather than n (VA_FORMAT_STRING_USES_NEWLINE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#fs-format-string-should-use-n-rather-than-n-va-format-string-uses-newline).
This format string includes a newline character (\n). In format strings, it is generally  preferable to use %n, which will produce the platform-specific line separator.

src/main/java/frc/robot/subsystems/FalconDriveSubsystem.java
FireShredder24: What if instead of polling Preferences and writing the values to the Talon FXs every scheduler cycle, we only did that once per second?  That would be trivially easy to implement and would basically accomplish the same goal.

terbos90803: It might help to understand the root problem.
The SparkMax docs say specifically that certain values like the PID settings shouldn't be written frequently.  I suspect the TalonFX has a similar warning.
The reason is because those values are held in a form of non-volatile memory (flash or eeprom) which has a limited number of write cycles in its lifetime.  You want to make sure you only write that when you're changing it.  Writing the same value repeatedly only wears out the storage and shortens the life of the device.

Theoretically, this type of check should be happening in the vendor code between you and the device, but you can never know for sure and if the docs say don't do it, then it's a safer bet to assume they aren't checking for you.

jordan-powers: There is a way to subscribe to Networktable entries. However, the WPILib Preferences class doesn't let us access those entries. Just the values.

PR #27, opened 2021-02-03
Add MoPrefsKey enum

jordan-powers:
MoPrefs are now accessed using the get() method, which takes a MoPrefsKey.
MoPrefsKeys are responsible for tracking the default value of their
associated preference, and are used as keys into the Preferences table.

This improves maintainability: adding a new preference only requires
adding a new entry to the MoPrefsKey enum instead of a whole new method
on MoPrefs.

See https://github.com/momentumfrc/2020InfiniteRecharge/pull/26#discussion_r568976978

### DIFF
--- a/src/main/java/frc/robot/utils/MoPrefs.java
+++ b/src/main/java/frc/robot/utils/MoPrefs.java
@@ -3,17 +3,57 @@ package frc.robot.utils;
 import edu.wpi.first.wpilibj.Preferences;
 
 public class MoPrefs {
-  static final double INTAKE_ROLLER_SETPOINT = 0.125;
-  static final double INTAKE_ROLLER_ACC_RAMP = 0.1;
-  static final int CLIMBER_ENCODER_LIMIT = 10;
-  static final double SHOOTER_HOOD_SETPOINT = 100;
-  static final double SHOOTER_HOOD_POSITION_TOLERANCE = 2;
-  static final double SHOOTER_GATE_SETPOINT = 1;
-  static final double SHOOTER_PID_SETPOINT = 4500;
-  static final double STORAGE_SPEED = 0.75;
-  static final double SHOOTER_FLYWHEEL_TOLERANCE = 100; // RPM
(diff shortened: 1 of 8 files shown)


jordan-powers:
I only converted about half of MoPrefs. No point in converting everything if we decide this is a bad approach.

src/main/java/frc/robot/utils/MoPrefs.java
```diff
@@ -3,17 +3,57 @@
...
+      return this.name();
+    }
+
+    public double getDefaultValue() {
+      return defaultValue;
+    }
+  }
 
   // PID constants for the shooter flywheel
   static final double SHOOTER_KP = 0.00005;
```
terbos90803: Why are these values still hanging out here?  Did you want to validate the approach before you finished all the work?

terbos90803 (commented):
Just one question.

src/main/java/frc/robot/utils/MoPrefs.java
jordan-powers: Pretty much. No point in putting in a lot of work converting the values if we decide we don't like this approach.

src/main/java/frc/robot/utils/MoPrefs.java
```diff
@@ -38,7 +38,43 @@
     // FIXME: Copy this value to the new key
     // The original key for this value was "Shoot From Wall Hood Setpoint"
     // The new key is "SHOOT_FROM_WALL_HOOD_SETPOINT"
-    SHOOT_FROM_WALL_HOOD_SETPOINT(60);
+    SHOOT_FROM_WALL_HOOD_SETPOINT(60),
+
```
terbos90803, line 42: Why all the extra blank lines?

jordan-powers: The automatic formatter tries to fit the declarations onto as few lines as possible. Adding the blank lines keeps each declaration on its own line. IMO, this is the more readable of the two options.

terbos90803 (approved):

